PR #1, opened 2022-04-20
Service base + pin endpoint

ro-tex:
# PULL REQUEST

## Overview

This PR initialises the service and adds the `/pin` endpoint which informs the pinner that a given skylink is being pinned by the local server.

The PR is large but it's mostly our standard boilerplate. I've kept the functionality to a minimum, so the review can be easy.

## Checklist
 - [ ] Changelog file created


test/api/handlers_test.go
```diff
@@ -0,0 +1,73 @@
...
+	// Pin an invalid skylink.
+	_, err := at.PinPOST("this is not a skylink")
+	if err == nil || !strings.Contains(err.Error(), database.ErrInvalidSkylink.Error()) {
+		t.Fatalf("Expected error '%s', got '%v'", database.ErrInvalidSkylink, err)
+	}
+	// Pin a valid skylink.
+	status, err := at.PinPOST(sl)
+	if err != nil || status != http.StatusNoContent {
+		t.Fatal(status, err)
+	}
```
MSevey, line 72: Follow Up: Will need tests for adding the same skylink for the same server, adding the same skylink for a different server.

MSevey (changes requested):

build/critical.go
```diff
@@ -0,0 +1,42 @@
...
+
+// buildInfoString is used to include information about the current build when
+// Critical or Severe are called.
+var buildInfoString = fmt.Sprintf("(Pinner v%v, Release: %v)", Version, Release)
+
+// Critical should be called if a sanity check has failed, indicating developer
+// error. Critical is called with an extended message guiding the user to the
+// issue tracker on Github. If the program does not panic, the call stack for
+// the running goroutine is printed to help determine the error.
+func Critical(v ...interface{}) {
```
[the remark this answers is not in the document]
ro-tex: ha, you're right, I can just pull it from skyd. thanks!

database/skylink.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// SkylinkCreate inserts a new skylink into the DB. Returns an error if it
+// already exists.
+func (db *DB) SkylinkCreate(ctx context.Context, sl string, server string) (Skylink, error) {
+	if !accdb.ValidSkylinkHash(sl) {
+		return Skylink{}, ErrInvalidSkylink
+	}
+	servers := []string{}
+	if server != "" {
```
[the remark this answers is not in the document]
ro-tex: Right, it would be better to just return an error because we shouldn't hit that case.

main.go
```diff
@@ -0,0 +1,51 @@
...
+		log.Fatal(errors.AddContext(err, "failed to connect to the db"))
+	}
+
+	// Initialise the server.
+	server, err := api.New(db, logger)
+	if err != nil {
+		log.Fatal(errors.AddContext(err, "failed to build the api"))
+	}
+
+	log.Fatal(server.ListenAndServe(4000))
```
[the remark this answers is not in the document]
ro-tex: I'm not sure it's worth it, tbh. We're running this in docker, which means that we don't face a risk of colliding port numbers. And if we want to expose it on another port we can always do that in the docker-compose file.

Also, we don't have that configurable in either accounts, blocker, or malware-scanner.

If we think this would be useful, I can add it in a F/U.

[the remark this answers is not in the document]
MSevey: ah this is the internal port. I was thinking with was going to be the docker referenced in the dockerfile.

nm then.

MSevey (approved):

test/tester.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	go func() {
+		_ = srv.ListenAndServe()
+	}()
+	go func() {
+		select {
+		case <-ctxWithCancel.Done():
+			_ = srv.Shutdown(context.TODO())
+		}
+	}()
```
ChrisSchinnerl, line 84: f/u This is not very graceful because the goroutines might just be terminated when the test ends without making sure the http server actually shut down. One way to do that gracefully is something like this

```go
tester.externShutdownDone = make(chan error)
go func() {
  tester.externShutdownErr = srv.ListenAndServe()
  close(tester.externShutdownDone)
}()
```

and then 
```go
func (t *Tester) Close() error {
  if err := t.staticSRV.Shutdown(someCtx); err != nil {
    return err
  }
  <-tester.externShutdownDone
  return tester.externShutdownErr
}
```

database/database.go
```diff
@@ -0,0 +1,160 @@
...
+func New(ctx context.Context, creds DBCredentials, logger *logrus.Logger) (*DB, error) {
+	return NewCustomDB(ctx, dbName, creds, logger)
+}
+
+// NewCustomDB creates a new database connection to a database with a custom name.
+func NewCustomDB(ctx context.Context, dbName string, creds DBCredentials, logger *logrus.Logger) (*DB, error) {
+	if ctx == nil {
+		return nil, errors.New("invalid context provided")
+	}
+	if logger == nil {
```
peterjan, line 67: here we error out while in other cases we init to a discard logger, I'm in favour of defaulting to a discard logger everywhere I think that's kinda nice

database/database.go
```diff
@@ -0,0 +1,160 @@
...
+		mongoWriteConcern,
+		mongoWriteConcernTimeout,
+	)
+}
+
+// ensureDBSchema checks that we have all collections and indexes we need and
+// creates them if needed.
+// See https://docs.mongodb.com/manual/indexes/
+// See https://docs.mongodb.com/manual/core/index-unique/
+func ensureDBSchema(ctx context.Context, db *mongo.Database, log *logrus.Logger) error {
```
peterjan, line 128: we might be at a point where we should seriously consider extracting all this database template code into a repo, that will also help us apply the same design choices because there are definitely 2-3 styles floating around now

database/schema.go
```diff
@@ -0,0 +1,26 @@
...
+// schema defines a mapping between a collection name and the indexes that
+// must exist for that collection.
+var schema = map[string][]mongo.IndexModel{
+	collSkylinks: {
+		{
+			Keys:    bson.D{{"skylink", 1}},
+			Options: options.Index().SetName("skylink").SetUnique(true),
+		},
+		{
+			Keys:    bson.D{{"lock_expires", 1}},
```
peterjan, line 18: what's this for?

database/skylink.go
```diff
@@ -0,0 +1,99 @@
...
+		ID      primitive.ObjectID `bson:"_id,omitempty"`
+		Skylink string             `bson:"skylink"`
+		Servers []string           `bson:"servers"`
+	}
+)
+
+// SkylinkCreate inserts a new skylink into the DB. Returns an error if it
+// already exists.
+func (db *DB) SkylinkCreate(ctx context.Context, sl string, server string) (Skylink, error) {
+	if !accdb.ValidSkylinkHash(sl) {
```
peterjan, line 38: kind of confused with the naming here, this will check whether it's a valid base32 or base64 encoded skylink string right? how is this different from the loadstring validation using the skylink type?

database/skylink.go
```diff
@@ -0,0 +1,99 @@
...
+		return Skylink{}, ErrInvalidSkylink
+	}
+	if server == "" {
+		return Skylink{}, errors.New("invalid server name")
+	}
+	s := Skylink{
+		Skylink: sl,
+		Servers: []string{server},
+	}
+	ir, err := db.staticDB.Collection(collSkylinks).InsertOne(ctx, s)
```
peterjan, line 48: wondering whether we should extend the `servers` array if the current server is not in there? so if you have a race condition you end up with a correct view of the skylink object rather than an error on one side? Just a thought.

database/skylink.go
```diff
@@ -0,0 +1,99 @@
...
+	return err
+}
+
+// SkylinkServerRemove removes a server to the list of servers known to be
+// pinning this skylink. If the skylink does not exist in the database it will
+// not be inserted.
+func (db *DB) SkylinkServerRemove(ctx context.Context, sl string, server string) error {
+	filter := bson.M{
+		"skylink": sl,
+		"servers": server,
```
peterjan, line 94: this is fine but it's not necessary, in some cases stuff like this can hurt your performance because your field is not indexed, although in this case it won't

skyd/skyd.go
```diff
@@ -0,0 +1,16 @@
+package skyd
+
+type (
+	// Client allows us to call the local skyd instance.
+	Client struct {
+	}
+)
+
+/*
+TODO
```
peterjan, line 10: as with the database, I feel like a skyd client like this is something we keep repeating across our repositories, we should probably at one point merge all of those together, or even better tweak skyd's client so we can use it in external repositories such as this one

test/api/handlers_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"strings"
+	"testing"
+
+	"github.com/skynetlabs/pinner/conf"
+	"github.com/skynetlabs/pinner/database"
+	"github.com/skynetlabs/pinner/test"
+	"gitlab.com/NebulousLabs/errors"
+)
+
+type subtest struct {
```
peterjan, line 14: `nit` docstring

test/api/handlers_test.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	dbName := test.DBNameForTest(t.Name())
+	at, err := test.NewTester(dbName)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		if errClose := at.Close(); errClose != nil {
+			t.Error(errors.AddContext(errClose, "failed to close account tester"))
```
peterjan, line 38: copy paste error msg

test/tester.go
```diff
@@ -0,0 +1,261 @@
...
+	// 64 characters is MongoDB's limit on database names.
+	// See https://docs.mongodb.com/manual/reference/limits/#mongodb-limit-Length-of-Database-Names
+	if len(cleanDBName) > 64 {
+		cleanDBName = cleanDBName[:64]
+	}
+	return cleanDBName
+}
+
+// Close performs a graceful shutdown of the Tester service.
+func (at *Tester) Close() error {
```
peterjan, line 126: `at` leftover

test/fixtures/mongo_keyfile
```diff
@@ -0,0 +1,16 @@
+PUQGGGC+g1sxoYzKKnX7531pos0WvDCYlx1Ak5ZupX3KvyuU2iVcC9puJJSG06VR
```
peterjan, line 1: I'm sure this is fine but I'm confused why we need a keyfile?

peterjan (commented):
Looking good! I guess we're close to merging this as a base to build on. Haven't clicked approved but all of my comments are fine to tackle in a F/U if you want to merge this in quick.

test/fixtures/mongo_keyfile
ro-tex: We want to use the same Mongo setup as in production. The initial idea was to be able to swap this for a staging server / database.

I'm fine with removing it if you're sure we don't need it. I set this up a long time ago and I haven't revisited it since then.

skyd/skyd.go
ro-tex: While I do agree that we should move the DB stuff to a common library, other clients do not necessarily fall into the same bucket. Different services use different `skyd` endpoints and make different assumptions in order to be as easy to use as possible. As such, a client developed for one service might be inadequate for another.

Also, the base implementation of the `skyd` client is still the client within the `skyd` project itself. As you can see in #2, the local client implementation is tiny and all the heavy lifting is done by `gitlab.com/SkynetLabs/skyd/node/api/client`.

database/skylink.go
ro-tex: I'm not sure what you mean. Is it that we filter by two values and not just one? I don't mind removing the filtering by `servers`.

Both fields are indexed, btw.

database/skylink.go
ro-tex: Makes sense, added.

database/skylink.go
ro-tex: replaced by `LoadString`.

database/schema.go
ro-tex: This service needs to do two things:
1. Track which server is pinning a given skylink.
2. Ensure that each skylink on record (that needs to be kept alive) is being pinned by a minimum number of servers.

To the second point, if a skylink is underpinned (in the sense that it's pinned by fewer than the minimum number of servers) it will be pinned by Pinner on some server. Since pinning takes some time, we want to indicate that the skylink is being currently pinned by a server, so other servers won't try to pin it as well. This is the purpose of the `lock` field. The `lock` field isn't implemented yet but it made sense for the `lock_expires` field to be added now, so we could use it when writing all mongo queries that we need today.

database/database.go
ro-tex: As we've already discussed, I'm on board with that. Let's discuss it in a separate place, though.

database/database.go
ro-tex: I'm very much against defaulting to a discard logger.
This is a big decision about the way the service reports its state and it should be made consciously and not by default. The only place where I would use a discard logger is during testing. Otherwise we're flying blind in production.

conf/configuration.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"os"
+
+	"github.com/joho/godotenv"
+	"gitlab.com/NebulousLabs/errors"
+)
+
+var (
+	// AccountsHost defines the IP or hostname of the local accounts service.
+	AccountsHost = "10.10.10.70"
```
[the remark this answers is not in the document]
ro-tex: This is a simple case of a default value. I find it more readable than 
```go
	if val, ok = os.LookupEnv("SKYNET_ACCOUNTS_HOST"); ok {
		AccountsHost = val
	} else {
		AccountsHost = "10.10.10.70"
        }
```

api/handlers.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// unpinPOST informs pinner that a given skylink should no longer be pinned by
+// any server.
+/*
+TODO
+ - Mark the skylink for unpinning.
+ - Unpin the skylink from the local server and remove the server from the list.
+ - Keep the skylink in the DB with the unpinning flag up. This will ensure that
+ if the skylink is still pinned to any server and we sweep that sever and add
```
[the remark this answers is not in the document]
ro-tex: ...

api/handlers.go
```diff
@@ -0,0 +1,85 @@
...
+
+// pinPOST informs pinner that a given skylink is pinned on the current server.
+func (api *API) pinPOST(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	var body SkylinkRequest
+	err := json.NewDecoder(req.Body).Decode(&body)
+	if err != nil {
+		api.WriteError(w, err, http.StatusBadRequest)
+		return
+	}
+	// Validate the skylink.
```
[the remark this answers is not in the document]
ro-tex: Good catch.

api/handlers.go
```diff
@@ -0,0 +1,85 @@
...
+	"gitlab.com/SkynetLabs/skyd/skymodules"
+)
+
+type (
+	// HealthGET is the response type of GET /health
+	HealthGET struct {
+		DBAlive bool `json:"dbAlive"`
+	}
+	// SkylinkRequest describes a request that only provides a skylink.
+	SkylinkRequest struct {
```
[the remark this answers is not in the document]
ro-tex: It is used by both `/pin` and `/unpin` endpoints. The fact that `/unpin` isn't in this PR doesn't change the master plan.
If we need to change one of those, we can always split them then. But duplicating code for the sake of a potential future need is an established WET anti-pattern.

test/utils.go
```diff
@@ -0,0 +1,66 @@
...
+var (
+	// TestServerName is what we'll use for ServerName during testing. We want
+	// to have it in a separate variable, so we can set it in different tests
+	// without worrying about them choosing different names.
+	TestServerName = "test.server.name"
+)
+
+type (
+	// ResponseWriter is a testing ResponseWriter implementation.
+	ResponseWriter struct {
```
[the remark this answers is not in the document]
ro-tex: I know of its existence but I haven't delved into it. I'll take a look but I'd rather do that in a F/U.

conf/configuration.go
```diff
@@ -0,0 +1,69 @@
...
+	// LogLevel defines the logging level of the entire service.
+	LogLevel = "info"
+	// ServerName holds the name of the current server. This name will be used
+	// for identifying which servers are pinning a given skylink.
+	ServerName string
+)
+
+// LoadConfiguration loads the required service configuration from the
+// environment and the provided .env file.
+func LoadConfiguration() error {
```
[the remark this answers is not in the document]
ro-tex: Hm... Yes, this will be nicer.

database/database.go
```diff
@@ -0,0 +1,160 @@
...
+	defer cancel()
+	return db.staticDB.Client().Ping(ctx2, readpref.Primary())
+}
+
+// connectionString is a helper that returns a valid MongoDB connection string
+// based on the passed credentials and a set of constants. The connection string
+// is using the standalone approach because the service is supposed to talk to
+// the replica set only via the local node.
+// See https://docs.mongodb.com/manual/reference/connection-string/
+func connectionString(creds DBCredentials) string {
```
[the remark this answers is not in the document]
ro-tex: Right, we discussed it about `malware-scanner` but we didn't apply to other services. I'll use it here and I'll add it to `accounts` as well with the next PR there.

main.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	// Initialise the global context and logger. These will be used throughout
+	// the service. Once the context is closed, any background threads will
+	// wind themselves down.
+	ctx := context.Background()
+	logger := logrus.New()
+	logLevel, err := logrus.ParseLevel(conf.LogLevel)
+	if err != nil {
+		logLevel = logrus.InfoLevel
```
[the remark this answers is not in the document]
ro-tex: Hm, good point. Initially, this was done to accommodate an empty value in the config because it's not an obligatory conf value. But now that we have a decent configuration initialisation, any error here will be caused by the config file and we should die on it.

test/tester.go
ro-tex: F/U: https://github.com/SkynetLabs/pinner/issues/4

conf/configuration.go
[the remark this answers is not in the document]
ChrisSchinnerl: Default vars are perfectly fine but the problem is overwriting them due to the races that might cause. So similar to the allowance in Sia we can do 

```go
// Config contains configuration for the pinner service.
type Config struct {...}

// DefaultConfig contains the defaults we use for the pinner service unless overwritten by env vars.
var DefaultConfig = Config {...}
```

and use it like this:

```go
func main() {
  // Load config.
  cfg, err := loadConfig()

  // Create service from config.
  service, err := foo.New(cfg)
}
```

test/database/skylink_test.go
```diff
@@ -0,0 +1,87 @@
...
+	// Validate that the underlying skylink is the same.
+	if s.Skylink != sl {
+		t.Fatalf("Expected skylink '%s', got '%s'", sl, s.Skylink)
+	}
+	// Add the skylink again, expect ErrSkylinkExists.
+	_, err = db.SkylinkCreate(ctx, sl, conf.ServerName)
+	if !errors.Contains(err, database.ErrSkylinkExists) {
+		t.Fatalf("Expected '%s', got '%v'", database.ErrSkylinkExists, err)
+	}
+	// TODO Add the same skylink again, this time in base32. Make sure it doesn't get duplicated.
```
ChrisSchinnerl, line 66: Did you mean to add this in this MR or f/u with it?

For reference we don't allow TODOs in skyd because we have come to the realisation that they clutter the codebase because we simply forget about them. 

This is the first MR for this repo so it's fine but in this case it's not clear whether you intended to add that or if it's just a leftover TODO from a different repo that we never implemented.

test/api/handlers_test.go
ChrisSchinnerl: We have been very strict about adding a docstring to every type, method and functions even when unexported because we have suffered a lot of confusion in the past (especially in testing).

For this type it's indeed pretty ridiculous to add a comment but you could also just inline it below since it appears to be the only time you are actually using it.

Either way it's not blocking this PR.

ChrisSchinnerl (dismissed):

test/database/skylink_test.go
ro-tex: The idea was to do it in a F/U. I actually ended up using skyd's `skylink.LoadString`, making this redundant.

test/api/handlers_test.go
ro-tex: I intend to use it in other tests in the future.
Added a docstring.

test/utils.go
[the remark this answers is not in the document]
ro-tex: https://github.com/SkynetLabs/pinner/issues/5

conf/configuration.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"os"
+
+	"github.com/joho/godotenv"
+	"gitlab.com/NebulousLabs/errors"
+)
+
+var (
+	// configuration is the current state of the configuration.
+	configuration = Config{
```
ChrisSchinnerl, line 12: f/u 

I think there was a misunderstanding. It's fine for this to be exposed as `DefaultConfiguration` for testing or whatever.
But we should never assign to global vars like we do in `LoadConf`.

Right now we got an unnecessary race condition in `LoadConf` when we call it from 2 threads because both are assigning to a global var. During a test which spins up 2 pinners this will fail with a race condition.

So we don't need `func Conf() Config`. Just a `func LoadConf() (Config, error) {` which creates a new `Conf` instance. That way we can call it a million times in parallel without the race detector going off because we are assigning to the same var.

ChrisSchinnerl (approved):

test/fixtures/mongo_keyfile
peterjan: 🤷‍♂️  I would remove it and use a simple config to set up the test db with. I think we should not necessarily make it easy to swap out staging/production config because that's quite risky actually and you can accomplish the same thing with just adjusting the URI, username and password.

MSevey (approved):

conf/configuration.go
ro-tex: https://github.com/SkynetLabs/pinner/issues/6

conf/configuration.go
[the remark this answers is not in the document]
ro-tex: https://github.com/SkynetLabs/pinner/issues/6

PR #2, opened 2022-04-22
SKY-146 Unpin endpoint

### LINKED ISSUE #6: Configuration race conditions
f/u 

I think there was a misunderstanding. It's fine for this to be exposed as `DefaultConfiguration` for testing or whatever.
But we should never assign to global vars like we do in `LoadConf`.

Right now we got an unnecessary race condition in `LoadConf` when we call it from 2 threads because both are assigning to a global var. During a test which spins up 2 pinners this will fail with a race condition.

So we don't need `func Conf() Config`. Just a `func LoadConf() (Config, error) {` which creates a new `Conf` instance. That way we can call it a million times in parallel without the race detector going off because we are assigning to the same var.

_Originally posted by @ChrisSchinnerl in https://github.com/SkynetLabs/pinner/pull/1#discussion_r859759389_

Also: https://github.com/SkynetLabs/pinner/pull/1#discussion_r858720075

ro-tex:
# PULL REQUEST

## Overview

This PR does the following:
* adds the unpin endpoint and introduces the `skyd` client
* introduces the concept of a skylink being marked as unpinned (i.e. needing to be removed by all servers)
* changes the way the service configuration works, so there are no data races

### Details

If a skylink is marked as unpinned and any server receives a `POST /pin` for it, that would immediately unmark the skylink as unpinned.

All of the pinning / unpinning of skylinks will be done by a background task added in https://github.com/SkynetLabs/pinner/pull/7.

There might be some additions to this PR which are not strictly necessary here because this one and #7 were developer together and then separated for easier reviewing.

## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes #6 #SKY-167


main.go
```diff
@@ -47,5 +47,17 @@ func main() {
 		log.Fatal(errors.AddContext(err, "failed to build the api"))
 	}
 
+	// TODO Start a background loop that would scan the DB for either unpinned
+	// 	or underpinned skylinks and pins/unpins them.
```
MSevey, line 51: we should make the `underpinned` value an environment variable so that we can increase/decrease the redundancy easily.

skyd/skyd.go
```diff
@@ -1,16 +1,133 @@
...
+	skylinksCache.mu.Lock()
+	defer skylinksCache.mu.Unlock()
+	// Check whether the cache is still valid and return it if so.
+	if skylinksCache.Expiration.After(time.Now().UTC()) {
+		return skylinksCache.Skylinks, nil
+	}
+	// The cache is not valid, fetch the data from skyd.
+	sls := make([]string, 0)
+	dirsToWalk := []skymodules.SiaPath{skymodules.SkynetFolder}
+	for len(dirsToWalk) > 0 {
```
MSevey, line 84: this should be moved to skyd.
We should have an endpoint to return all skylinks.

test/api/handlers_test.go
```diff
@@ -80,4 +81,51 @@ func testHandlerPinPOST(t *testing.T, at *test.Tester) {
...
 	}
+
+	// Mark the skylink as unpinned and pin it again.
+	// Expect it to no longer be unpinned.
+	err = at.DB.SkylinkMarkUnpinned(at.Ctx, sl)
+	if err != nil {
+		t.Fatal(err)
+	}
+	status, err = at.PinPOST(sl)
+	if err != nil || status != http.StatusNoContent {
```
MSevey, line 92: why is `http.StatusNoContent` an acceptable response here?

test/api/handlers_test.go
```diff
@@ -80,4 +81,51 @@ func testHandlerPinPOST(t *testing.T, at *test.Tester) {
...
+		t.Fatal(status, err)
+	}
+	// Make sure the skylink is marked as unpinned.
+	slNew, err := at.DB.SkylinkFetch(at.Ctx, sl)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !slNew.Unpin {
+		t.Fatal("Expected the skylink to be marked as unpinned.")
+	}
```
MSevey, line 130: I would also add calling unpin on a valid skylink that doesn't exist to verify that behavior.

test/api/handlers_test.go
ro-tex: `StatusNoContent` (204) is `skyd`'s default success code and we're using it across all services as such.
`WriteSuccess()` writes a 204.

skyd/skyd.go
ro-tex: This can be done in a F/U after the functionality is added to `skyd`. For the moment I would prefer to move as fast as possible and go ahead with this approach just so we can start pinning.

skyd issue: https://gitlab.com/SkynetLabs/skyd/-/issues/328

main.go
ro-tex: Hmm... This should be a cluster-wide setting. A DB-based configuration will make more sense, so we can adjust the entire cluster's pinning level simultaneously. Otherwise some servers might think that 1x is enough while others try to get 2x, thus taking way more load onto themselves than they should.

test/api/handlers_test.go
ro-tex: Done.

api/api.go
```diff
@@ -27,7 +29,7 @@ type (
 )
 
 // New returns a new initialised API.
-func New(db *database.DB, logger *logrus.Logger) (*API, error) {
+func New(cfg conf.Config, db *database.DB, logger *logrus.Logger) (*API, error) {
```
peterjan, line 32: we don't generally inject high level config structs that contain all configuration into services but instead only supply the service with the information it needs and nothing more

peterjan: I don't really mind it's just that doing it this way it becomes harder to see what services require what configuration values.

conf/configuration.go
```diff
@@ -29,60 +30,79 @@ type (
 	// LogLevel defines the logging level of the entire service.
 	// ServerName holds the name of the current server. This name will be used
 	// 	for identifying which servers are pinning a given skylink.
+	// SiaAPIPassword is the apipassword for the local skyd
+	// SiaAPIHost is the hostname/IP of the local skyd
+	// SiaAPIPort is the port of the local skyd
```
peterjan, line 35: why are these descriptions not in-line with the struct like we normally do, now I have to scan the docstring for an explanation of `SiaAPIPort` which is 10 lines up and I have to also know to consult the docstring for that info 😛  I sound like the most anal person ever saying that but I don't see the benefit of this new way of documenting struct fields

database/skylink.go
```diff
@@ -21,22 +22,31 @@ var (
 	// ErrSkylinkNoExist is returned when we try to get a skylink that doesn't
 	// exist.
 	ErrSkylinkNoExist = errors.New("skylink does not exist")
+
+	// LockDuration defines the duration of a database lock. We lock skylinks
+	// while we are trying to pin them to a new server. The goal is to only
+	// allow a single server to pin a given skylink at a time.
+	LockDuration = 24 * time.Hour
```
peterjan, line 29: why not re-use the `mongo-lock` dependency we have in several places, collection already exists just have to pick a unique resource name

peterjan: these locks are currently not implemented yet right? couldn't find much references to it

database/skylink.go
```diff
@@ -72,34 +82,83 @@ func (db *DB) SkylinkFetch(ctx context.Context, sl string) (Skylink, error) {
...
-	_, err := db.staticDB.Collection(collSkylinks).UpdateOne(ctx, filter, update, &opts)
+	_, err = db.staticDB.Collection(collSkylinks).UpdateOne(ctx, filter, update, &opts)
 	return err
 }
 
 // SkylinkServerRemove removes a server to the list of servers known to be
 // pinning this skylink. If the skylink does not exist in the database it will
 // not be inserted.
-func (db *DB) SkylinkServerRemove(ctx context.Context, sl string, server string) error {
+func (db *DB) SkylinkServerRemove(ctx context.Context, skylink string, server string) error {
```
peterjan, line 141: you could have the database simply handle `skymodules.Skylink` and take care of the validation higher up, you can resolve this comment if you feel that's not a good idea was just suggesting it because it kind of simplifies things I think (maybe) to do this in the handler itself.

E.g. 
```go
sl, err := validateSkylink(params.Skylink)
if err != nil { return 400 } 
err = db.skylinkRemove(sl)
if err != nil { return 500 } 
```
rather than
```go
err := db.skylinkRemove(params.Skylink)
if err != nil {
    if errors.Contains(err, ErrInvalidSkylink) {
        return 400
    }
    return 500
}
```

database/skylink.go
```diff
@@ -72,34 +82,83 @@ func (db *DB) SkylinkFetch(ctx context.Context, sl string) (Skylink, error) {
...
+// SkylinkMarkUnpinned marks a skylink as unpinned, meaning that all servers
+// should stop pinning it.
+func (db *DB) SkylinkMarkUnpinned(ctx context.Context, skylink string) error {
+	sl, err := CanonicalSkylink(skylink)
+	if err != nil {
+		return ErrInvalidSkylink
+	}
+	filter := bson.M{"skylink": sl}
+	update := bson.M{"$set": bson.M{"unpin": true}}
+	opts := options.UpdateOptions{}
```
peterjan, line 116: just in case you didn't know about this, I don't mind the current format
`opts := options.Update().SetUpsert(true)`

database/skylink.go
```diff
@@ -72,34 +82,83 @@ func (db *DB) SkylinkFetch(ctx context.Context, sl string) (Skylink, error) {
...
+	opts.SetUpsert(true)
```
peterjan, line 117: this is probably ok but just checking in case this is  a copy-paste mistake, do you want to upsert on unpinning also?

peterjan (commented):
LGTM over all, not sure why we're not using our locking lib and then just have a couple of minor remarks, other than that looks solid 👍

database/skylink.go
peterjan: seems a bit high as well? is there a reasoning behind it being 24h?

api/api.go
ro-tex: that's reasonable. changed to pass only what a component needs.

conf/configuration.go
ro-tex: having comments on each field of a struct makes it very hard for me to read, so I moved those to the overall description of the type. reverted, the comments are now per-field.

database/skylink.go
ro-tex: 1. Using the locking library makes sense, I'll look into it.
2. Yes, the implementation is in #7. I split the full implementation into two PRs and some stuff has ended up in the first one while only being needed in the second.
3. 24 hours is fine because we want to scan these once per day at most. Once a skylink is locked by a certain server, it will be the first thing that server pulls from the DB until it's unlocked. So, if the server's execution terminates, it will pick it right back up.

database/skylink.go
ro-tex: Yes. The reason being that we mark a skylink as unpinned when we actively want to stop pinning in on all of our servers. The motivation for that would be that no user is currently pinning that skylink. So, if the skylink doesn't exist in our DB we want to create it in order to notify all servers that they need to remove it.

database/skylink.go
ro-tex: yes, I know about it and I'm suing it but it seems I'm inconsistent with it :)

btw, a nice side benefit of this approach is that if you forget to pass the `opts` to the actual `Update` your editor will mark it as unused while if you do `opts.SetUpsert(true)` it won't.

database/skylink.go
ro-tex: That sounds good. It seems like you don't necessarily need it in this PR, so I'd rather add it to #7 in order to keep the conflicts down.

database/skylink.go
ro-tex: I'll actually add the locking library in #7 where the locks are actually used.

test/api/handlers_test.go
MSevey: no i get it now.

MSevey (dismissed):

main.go
MSevey: agree that it is a cluster wide variable, but it should still be an env var so we don't need to submit a pull request and deploy to change it. 

follow up, just calling it out.

api/handlers.go
```diff
@@ -38,14 +39,29 @@ func (api *API) pinPOST(w http.ResponseWriter, req *http.Request, _ httprouter.P
...
 	}
 	// Create the skylink.
-	_, err = api.staticDB.SkylinkCreate(req.Context(), body.Skylink, conf.Conf().ServerName)
+	_, err = api.staticDB.SkylinkCreate(req.Context(), body.Skylink, api.staticConfig.ServerName)
 	if errors.Contains(err, database.ErrInvalidSkylink) {
 		api.WriteError(w, err, http.StatusBadRequest)
 		return
 	}
 	// If the skylink already exists, add this server to its list of servers.
 	if errors.Contains(err, database.ErrSkylinkExists) {
```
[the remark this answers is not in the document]
ro-tex: Here we can technically call `SkylinkServerAdd` and leave it at that. The issue with that is that if we add other fields to the skylink record we won't be setting it there. But that that point we'll need to remove the upsert flag as well.

More specifically to your non-acid vibes, I believe the issue before was that I was doing a non-acid sequence:
1. fetch the list
2. add/remove item
3. store the list

But this time I'm not doing that and instead I'm using `update := bson.M{"$addToSet": bson.M{"servers": server}}` which should be acid.

database/skylink.go
```diff
@@ -63,7 +69,11 @@ func (db *DB) SkylinkCreate(ctx context.Context, sl string, server string) (Skyl
 }
 
 // SkylinkFetch fetches a skylink from the DB.
-func (db *DB) SkylinkFetch(ctx context.Context, sl string) (Skylink, error) {
+func (db *DB) SkylinkFetch(ctx context.Context, skylink string) (Skylink, error) {
```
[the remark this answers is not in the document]
ro-tex: It leaves quite a bit of ambiguity about the behaviour. Does it fetch one? Does it create it? You can't easily tell at a glance without knowing how the DB module works.

I don't mind renaming it, if you'd like it better. I'd prefer to do it in #7, if we're going it.

api/handlers.go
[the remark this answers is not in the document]
peterjan: This is just missing distributed resource locking. By adding that you can easily make this ACID right.

database/skylink.go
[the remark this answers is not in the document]
peterjan: I usually model the database interface very closely to the underlying driver or database technology. So in this case I would go for `FindSkylink`. 

I think I agree with Ivo when it comes to just calling it `Skylink`. I would not be sure whether it would create the Skylink if it does not exist, or if I would get a Skylink object that hasn't been persisted yet, or maybe something else even.

I don't like that `Fetch` comes after `Skylink`, dogs fetch 😛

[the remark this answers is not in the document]
ro-tex: The reason it's `Skylink<action>` is so you can type `db.Skyl` and press your auto-complete shortcut and get a list of all available actions on skylinks.

Regarding the term - in DB systems `fetch` and `query` are used more or less interchangeably. MySQL and Postgres use `fetch`, DynamoDB uses `query`, and MongoDB apparently uses `find`. I guess what we find more logical depends on which DB we started with or used for longest time. I don't mind renaming this to `Find` but I typically expect `find` operations to return a list and not a single result. I'm not naming this `FindOneSkylink`, though.

api/handlers.go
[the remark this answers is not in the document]
ro-tex: Do we need distributed resource locking here, though? We have two operations that we attempt - create and update.

The creation shouldn't need resource locking because if there is a collision we know that only one of the attempts will succeed and all others will fail, which is handled by the code.

The update should also be fine because, afaik, the `$addToSet` operation is atomic, so it doesn't need a lock.

database/skylink.go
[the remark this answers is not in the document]
ChrisSchinnerl: I don't think there would be a ton of confusion because that's just idiomatic getter/setter naming in Go. You want a skylink call it `Skylink`. You want to set one you call `SetSkylink`. You want to create one call `NewSkylink`.

If you'd like to put more emphasis on the fact that it involves I/O from a Mongo database then I'd go straight for `FindSkylink` to group all the operations that fetch data together and also to adhere to the mongo naming conventions. I generally find it more useful to group by action when using a library I don't know yet.
One example would be `binary.LittleEndian.PutUint64` / `binary.LittleEndian.Uint64`. Seems easier for me to type the operation I want and then see the variations than the other way round.

api/handlers.go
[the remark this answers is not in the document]
ChrisSchinnerl: The problem is that you could do `SkylinkCreate` and it tells you the skylink exists.
Then you call `SkylinkFetch` but in the meantime the skylink was deleted so it will fail and it will write a `500` to the API in line 52.

Similar thing if it happens after the fetch. You might call `SkylinkMarkPinned` or `SkylinkSetverAdd` on a link that doesn't exist and return a 500 which is not ideal.

I wouldn't introduce locking though if there is a way around it. e.g. by doing a `findAndModify` with upsert or something like that.

ChrisSchinnerl (dismissed):

api/handlers.go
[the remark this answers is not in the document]
ro-tex: From offline:

We don't need the fetch if we add a flag for the `SkylinkServerAdd` method which sets the `Unpin` field to `false` if the method is being called because a user pinned a file. The alternative case for calling this method would be when performing a `sweep` action, which takes all skylinks pinned by skyd and registering them in the DB.  
It seems that removing the fetch step cleans up the code enough and makes most sense, so I'll go with that.

database/skylink.go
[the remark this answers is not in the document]
ro-tex: We seem to be discussing two separate things here, so let's lay them out.

1. Order of "action" and "subject" when naming, e.g. "CreateSkylink" vs "SkylinkCreate".

My argument for "<subject><action>" is the ease of using auto-complete to list all available actions on a given subject. This is helpful for people who are unfamiliar with the service and want to achieve something.

That said, if we have stronger support for "<action><subject>" I won't block that and I'll refactor the service. I just find `AddServerSkylink`/`ServerAddSkylink`/`AddSkylinkServer` to be awkward and confusing - I will need to consult the docs before using that function because the name is not clear enough.

2. Action names - `set`/`new`/`create`.

If we want to go for standard go naming (new, get, set, update), I can agree to that. I'm not sure it always makes perfect sense when dealing with external data stores but it's acceptable.

I wouldn't go for Mongo-specific method naming because underlying technologies such as DB implementation should be replaceable without rewriting more than the persistence layer of the service. I don't consider either `find` or `fetch` to be technology specific in this sense, what I have in mind here is more stuff like `findOne` and `findAndUpdate`.

In any case, this doesn't seem like a blocking discussion and it also seems like a long conversation that should happen on Discord, rather than GitHub. To that end, I'll open a new thread on `#core-dev` and once we have a decision I'll refactor the service accordingly.

ChrisSchinnerl (approved):

MSevey (approved):

peterjan (approved):

PR #14, opened 2022-05-10
Boilerplate

ro-tex:
# PULL REQUEST

## Overview

Add Dockerfile.
Add Changelog.
Add GitHub actions.

## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes SKY-150

### DIFF
--- a/.github/workflows/ci_release.yml
+++ b/.github/workflows/ci_release.yml
@@ -1,0 +1,71 @@
+name: Release
+on:
+  push:
+    # Nightly schedule pending full e2e testing with other repos
+    #schedule:
+    # Run daily at 1:15am
+    #- cron: "15 1 * * *"
+  workflow_dispatch:
+    # Inputs the workflow accepts.
+    inputs:
+      version:
+        # Friendly description to be shown in the UI instead of 'name'
+        description: "Semver type of new version (major / minor / patch)"
+        # Input has to be provided for the workflow to run
+        required: true
+        type: choice
+        options:
+          - patch
+          - minor
+          - major
+
+jobs:
+  # Run the linting and tests
+  hadolint:
+    uses: SkynetLabs/.github/.github/workflows/reusable_dockerfile_lint.yml@master
+
+  test:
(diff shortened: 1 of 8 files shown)


.github/workflows/ci_release.yml
```diff
@@ -0,0 +1,71 @@
...
+    uses: SkynetLabs/.github/.github/workflows/reusable_dockerfile_lint.yml@master
+
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "1.18"
```
MSevey, line 34: Follow Up: we should refactor a basic golang repo action.

changelog/README.md
```diff
@@ -0,0 +1,4 @@
+# Changelog Generator
+
+For usage of changelog generator please see [readme](https://gitlab.com/NebulousLabs/changelog-generator/-/blob/master/README.md) in Changelog Generator
```
MSevey, line 3: Follow Up for me, we should trigger this from the github actions on release.

MSevey: added to linear.

.github/workflows/ci_release.yml
MSevey: added to linear

MSevey (approved):

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.18.1 as builder
```
MSevey, line 1: Follow Up: we should have a base dockerfile that sets the golang version so that we ensure all golang repos are using the same golang version.

kwypchlo: I don't think that's a good idea, we should let every image decide the version (but keep it updated with dependabot) otherwise we would be vulnerable to breaking something when upgrading the version across all repos without properly testing them.

kwypchlo (approved):

PR #15, opened 2022-05-11
Follow ups from the initial scanner implementation.

### LINKED ISSUE #12: follow up; I would add a `parseSkylink` helper to ensure all endpoints and future endpoint parse the skylink the same way.
follow up; I would add a `parseSkylink` helper to ensure all endpoints and future endpoint parse the skylink the same way.

_Originally posted by @MSevey in https://github.com/SkynetLabs/pinner/pull/7#discussion_r868296372_

ro-tex:
# PULL REQUEST

## Overview

* Fix the scanner loop. 
* Add diagnostic logging.
* Add parse helper.

## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes #12

### DIFF
--- a/database/skylink.go
+++ b/database/skylink.go
@@ -86,6 +86,8 @@ func (db *DB) FindSkylink(ctx context.Context, skylink skymodules.Skylink) (Skyl
 // MarkPinned marks a skylink as pinned (or no longer unpinned), meaning
 // that Pinner should make sure it's pinned by the minimum number of servers.
 func (db *DB) MarkPinned(ctx context.Context, skylink skymodules.Skylink) error {
+	db.staticLogger.Tracef("Entering MarkPinned. Skylink: '%s'", skylink)
+	defer db.staticLogger.Trace("Exiting MarkPinned.")
 	filter := bson.M{"skylink": skylink.String()}
 	update := bson.M{"$set": bson.M{"unpin": false}}
 	opts := options.Update().SetUpsert(true)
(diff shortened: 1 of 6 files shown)


database/skylink.go
```diff
@@ -86,6 +86,8 @@ func (db *DB) FindSkylink(ctx context.Context, skylink skymodules.Skylink) (Skyl
 // MarkPinned marks a skylink as pinned (or no longer unpinned), meaning
 // that Pinner should make sure it's pinned by the minimum number of servers.
 func (db *DB) MarkPinned(ctx context.Context, skylink skymodules.Skylink) error {
+	db.staticLogger.Tracef("Entering MarkPinned. Skylink: '%s'", skylink)
+	defer db.staticLogger.Trace("Exiting MarkPinned.")
```
MSevey, line 90: Should these defer statements also have the skylink? So we can confirm which exit statement pertains to which skylink?

skyd/skyd.go
```diff
@@ -121,6 +139,9 @@ func (c *client) PinnedSkylinks() (map[string]interface{}, error) {
 		}
 		// Grab all subdirs and queue them for walking.
 		for _, d := range rd.Directories {
+			if d.SiaPath.String() == dir.String() {
```
MSevey, line 142: For context, the current directory is always the 0 index of the slice of returned directories. 

Just for future reference of why this came up.

MSevey (approved):

peterjan (approved):

database/skylink.go
ro-tex: Can do. It's not strictly necessary because we are only processing one skylink at a time but it doesn't hurt and it also would make it clearer if we start processing more than one skylink in parallel. I'll add it in a F/U/

skyd/skyd.go
ro-tex: I'll clean it up in the next F/U.

PR #17, opened 2022-05-11
Min Pinners

### LINKED ISSUE #8: SKY-279 Cluster-wide setting for underpinned threshold
we should make the `underpinned` value an environment variable so that we can increase/decrease the redundancy easily.

_Originally posted by @MSevey in https://github.com/SkynetLabs/pinner/pull/2#discussion_r856472538_

## Proposal

As we want this to be a cluster-wide setting, we cannot use an environment variable for it because each server will have its own value and they will need to do a complicated negotiation in order to decide on a common, cluster-wide number.

That's why we should add a `configuration` or `settings` collection in Mongo and use that to store all cluster-wide settings, similar to what we do in `accounts` when we need to disable registrations on a given cluster.

ro-tex:
# PULL REQUEST

## Overview

Add a cluster-wide configuration field defining the minimum number of pinners we want.
In the process of doing that, this PR also defines a DB collection for cluster-wide configuration values and a generic way for getting those.

## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes #8 
Closes SKY-279


MSevey (dismissed):

conf/configuration.go
```diff
@@ -107,3 +118,25 @@ func LoadConfig() (Config, error) {
...
+		return defaultMinPinners, nil
+	}
+	if err != nil {
+		return 0, err
+	}
+	mp, err := strconv.ParseInt(val, 10, 0)
+	if err != nil {
+		return 0, err
+	}
+	if mp < 1 || mp > 10 {
```
ChrisSchinnerl, line 136: f/u I'd pull this out into a const with a quick explanation on why we limit it to 10.

workers/scanner.go
```diff
@@ -132,6 +136,7 @@ func (s *Scanner) threadedScanAndPin() {
 
 	// Main execution loop, goes on forever while the service is running.
 	for {
+		s.refreshMinPinners()
```
ChrisSchinnerl, line 139: This is not a requirement for this MR but something we should keep in the back of our heads is exploring mongo change streams (https://www.mongodb.com/docs/manual/changeStreams/#watch-a-collection--database--or-deployment).

If we ever have settings which we want to apply asap, that might be a good way to do it without polling.

ChrisSchinnerl (approved):

workers/scanner.go
ro-tex: This looks nice. I'll put it in a F/U.

conf/configuration.go
ro-tex: By the way, the limit of 10 is purely arbitrary. If you have suggestion for another limit, I'm happy to change the value. It's just a sanity check.

MSevey (approved):

conf/configuration.go
ro-tex: Since I needed to resolve merge conflicts anyway I just fixed this here.

MSevey (approved):

ChrisSchinnerl (approved):

PR #18, opened 2022-05-11
Rebuild the pinned skylinks cache once before each scan.

### LINKED ISSUE #11: you're holding this lock while doing a bunch of I/O, are you sure that's fine?
you're holding this lock while doing a bunch of I/O, are you sure that's fine?

_Originally posted by @peterjan in https://github.com/SkynetLabs/pinner/pull/7#discussion_r865697163_

ro-tex:
# PULL REQUEST

## Overview

This PR changes the way we manage the cached list of pinned skylinks.

#### Before
We used to get the list and consider it valid for an arbitrary period of time (one hour) after which period we would refresh it. 

#### Now
The new approach is simpler - we rebuild the cache at the start of each scanning cycle and we consider it valid for the duration of the cycle.

## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes #11
Closes SKY-375


skyd/skyd.go
```diff
@@ -98,11 +99,11 @@ func (c *client) Pin(skylink string) (skymodules.SiaPath, error) {
 func (c *client) PinnedSkylinks() (map[string]interface{}, error) {
 	skylinksCache.mu.Lock()
 	defer skylinksCache.mu.Unlock()
-	// Check whether the cache is still valid and return it if so.
-	if skylinksCache.expiration.After(time.Now().UTC()) {
```
MSevey, line 102: REQ: This check got dropped. So we are always rebuilding the cache every call regardless of the expiration.

MSevey (changes requested):

skyd/skyd.go
ro-tex: No, we are never rebuilding the cache here. The cache's expiration is irrelevant right now. I thought of removing it but didn't.

The way we operate right now is:
1. Rebuild the cache at the start of each scan.
2. Consider the cache valid until the end of the scan, regardless of its length.
3. Go to 1.

Based on the assumptions being made, this is a safe and simple implementation which has minimal risk of blowing up in our faces.

ChrisSchinnerl (approved):

MSevey (approved):

PR #19, opened 2022-05-11
Clearer errors

### LINKED ISSUE #9: This is saying that there are no skylinks that can be pinned by this server right?  Feels like we should just return the `mongo.ErrNoDocuments` in that case.
This is saying that there are no skylinks that can be pinned by this server right?  Feels like we should just return the `mongo.ErrNoDocuments` in that case. 

`ErrSkylinkNoExist` should only be used if we are trying to update a skylink record and that record doesn't exist.

_Originally posted by @MSevey in https://github.com/SkynetLabs/pinner/pull/7#discussion_r865146327_

ro-tex:
# PULL REQUEST

## Overview

This PR introduces a specific error for the case when there are no skylnks that need to be pinned by the current server.
It also adds a helper method for checking for that error.

## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes #9 
Closes SKY-373


test/database/skylink_test.go
```diff
@@ -165,8 +165,8 @@ func TestFindAndLock(t *testing.T) {
 
 	// Try to fetch an underpinned skylink, expect none to be found.
 	_, err = db.FindAndLockUnderpinned(ctx, cfg.ServerName, cfg.MinPinners)
-	if !errors.Contains(err, database.ErrSkylinkNotExist) {
-		t.Fatalf("Expected to get '%v', got '%v'", database.ErrSkylinkNotExist, err)
+	if !errors.Contains(err, database.ErrNoUnderpinnedSkylinks) {
```
MSevey, line 168: nit: don't you want to use your helper method here

MSevey (approved):

ChrisSchinnerl (approved):

test/database/skylink_test.go
peterjan: I wasn't going to `nit` on it but while scrolling through your MR I was going to remark that I think the helper is weirdly named, if you just follow the convention `isErrX` which then checks if the given error contains `errX` then this wouldn't have been confusing.

So I think renaming the helper to `IsErrNoUnderpinnedSkylinks` and using it here might be the F/U. But feel free to resolve it's a `nit`.

peterjan (approved):

test/database/skylink_test.go
ro-tex: Oh, sure, I'll rename it.

PR #21, opened 2022-05-16
Add docker builds and fix release github action

MSevey:
# PULL REQUEST

## Overview
Adds docker build GitHub action and Fix release action.

## Issues Closed

Closes SKY-470


.github/workflows/ci_release.yml
```diff
@@ -62,10 +62,28 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+
+      - uses: SkynetLabs/.github/.github/actions/docker-publish@master
+        with:
+          docker_username: ${{ secrets.DOCKER_USERNAME}}
+          docker_password: ${{ secrets.DOCKER_PASSWORD}}
+          docker_repository: skynetlabs/skynet-accounts
```
ro-tex, line 88:
```suggestion
          docker_repository: skynetlabs/pinner
```

ro-tex (dismissed):
LGTM.

.github/workflows/ci_release.yml
```diff
@@ -62,10 +62,28 @@ jobs:
     needs: changes
     runs-on: ubuntu-latest
     if: ${{ (needs.changes.outputs.updates == 'true' && github.event_name == 'schedule') || github.event_name == 'workflow_dispatch' }}
+    outputs:
+      new_version: ${{ steps.version.outputs.new-version }}
```
kwypchlo, line 66: have you checked whether dashed property works with dot accessor ? in javascript that would not work and you would need to do sth like `steps.version.outputs["new-version"]` so I'm just concerned and you're also assigning it to a variable with a underscore which also raises my concern :)

kwypchlo (changes requested):

.github/workflows/ci_release.yml
MSevey: yup, this was all tested on the docker test repo. All these PRs are not just copying the same code to all the repos. 

the `-` vs `_` doesn't matter. That is just mis matched conversations between actions. 

I tried to stay consistent in the actions themselves with how the external actions are set up. But then that leads to the inevitable mismatch for things like this. Ultimately there is no way around it, there will be mismatches somewhere. 

yml doesn't care if the variable has a `-` or a `_`

.github/workflows/ci_release.yml
```diff
@@ -62,10 +62,28 @@ jobs:
...
+          version-bump: ${{github.event.inputs.version}}
+
+  docker:
+    needs: release
+    name: Build and Push Docker image to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+
```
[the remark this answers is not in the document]
MSevey: you are right, we can remove this duplicate.

ro-tex (approved):

Issue #3, opened 2022-04-25
Get the list of pinned skylinks directly from skyd


ro-tex:
**NOTE**: This issue depends on https://gitlab.com/SkynetLabs/skyd/-/issues/328 being implemented first.

this should be moved to skyd.
We should have an endpoint to return all skylinks.

_Originally posted by @MSevey in https://github.com/SkynetLabs/pinner/pull/2#discussion_r856474174_

PR #22, opened 2022-05-17
Implement `POST /sweep`

ro-tex:
# PULL REQUEST

## Overview

`POST /sweep` aims to synchronise the content of the database with the list of files pinned by `skyd`.

The way to achieve that is to fetch both lists and calculate the diff between them. After that we iterate over the skylinks pinned by `skyd` but not present in the database records and we add them there. Then we iterate over all skylinks present in the database (under this server) and not pinned by `skyd` and we remove this server's name as their pinner in the database.

### Open questions

1. Currently, this is a manual operation. Should we perform this regularly?

## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes SKY-298


api/handlers.go
```diff
@@ -89,12 +91,67 @@ func (api *API) unpinPOST(w http.ResponseWriter, req *http.Request, _ httprouter
...
+	}
+	wg.Wait()
+	if cacheErr != nil {
+		api.WriteError(w, cacheErr, http.StatusInternalServerError)
+		return
+	}
+
+	unknown, missing := api.staticSkydClient.DiffPinnedSkylinks(dbSkylinks)
+
+	ctx := req.Context()
```
ro-tex, line 122: This is potentially a very slow operation. The context is almost guaranteed to time out.

Options:
1. Use a non-expiring background context.
2. Use a custom context with its own deadline which makes sense for this case.
3. Start the process in a separate goroutine (with a global semaphore which ensures only one sweep can be performed at a time) and respond with 202 Accepted. Optionally, we can provide a separate endpoint on which the status of the sweep can be monitored.

skyd/cache.go
```diff
@@ -0,0 +1,144 @@
...
+	// skylinksCache is a singleton cache, initialised on module load.
+	skylinksCache *pinnedSkylinksCache
+)
+
+type (
+	// pinnedSkylinksCache is a simple cache of the renter's directory
+	// information, so we don't need to fetch that for each skylink we
+	// potentially want to pin/unpin.
+	pinnedSkylinksCache struct {
+		rebuildCh chan interface{}
```
ChrisSchinnerl, line 26: The more common way to do this is `chan struct{}` and then if you ever need to push to it do `c <- struct{}{}`.

skyd/cache.go
```diff
@@ -0,0 +1,144 @@
...
+var (
+	// ErrNoRebuildInProgress is returned when we try to finish a rebuild but
+	// none is in progress.
+	ErrNoRebuildInProgress = errors.New("no rebuild in progress")
+	// ErrRebuildInProgress is returned when we try to start a rebuild and one
+	// is already in progress.
+	ErrRebuildInProgress = errors.New("rebuild in progress")
+
+	// skylinksCache is a singleton cache, initialised on module load.
+	skylinksCache *pinnedSkylinksCache
```
ChrisSchinnerl, line 18: Singletons are not great unless there is a very good reason to use them.
Can't we just create the cache with `newCache` and assign it to whatever is using it?

skyd/cache.go
```diff
@@ -0,0 +1,144 @@
...
+		mu:        sync.Mutex{},
+	}
+}
+
+// Add registers the given skylink in the cache.
+func (psc *pinnedSkylinksCache) Add(skylink string) {
+	psc.mu.Lock()
+	defer psc.mu.Unlock()
+	if psc.skylinks == nil {
+		psc.skylinks = make(map[string]interface{})
```
ChrisSchinnerl, line 48: This could happen in `init` I think.

skyd/cache.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	// Transform the removed map into a list.
+	for sl := range removedMap {
+		missing = append(missing, sl)
+	}
+	return
+}
+
+// Remove registers the given skylink in the cache.
+func (psc *pinnedSkylinksCache) Remove(skylink string) {
```
ChrisSchinnerl, line 90: Not sure it would be useful but since I assume that you often remove a lot of links at once, you could make this `Remove(skylink ...string)`. 

Then you can remove a batch while only locking/unlocking once.

skyd/cache.go
```diff
@@ -0,0 +1,144 @@
...
+	defer psc.mu.Unlock()
+	if psc.skylinks == nil {
+		psc.skylinks = make(map[string]interface{})
+	}
+	delete(psc.skylinks, skylink)
+}
+
+// blockingWaitForRebuild blocks until the current cache rebuild process ends.
+func (psc *pinnedSkylinksCache) blockingWaitForRebuild() {
+	<-psc.rebuildCh
```
ChrisSchinnerl, line 101: This might be a bit dangerous. Is this guaranteed to be unblocked?
Might be better to return the channel directly and then the caller can `select` on it.

skyd/cache.go
```diff
@@ -0,0 +1,144 @@
...
+	select {
+	case <-psc.rebuildCh:
+		return false
+	default:
+		return true
+	}
+}
+
+// managedReplaceCache replaces the entire cache content.
+func (psc *pinnedSkylinksCache) managedReplaceCache(newCache map[string]interface{}) {
```
ChrisSchinnerl, line 118: It seems a bit odd that the cache is replaced from an outsides source and also that an outside source is signaling a rebuild start and end. Vs the cache being able to rebuild itself and the caller simply calling `cache.Rebuild` on it.

api/handlers.go
ro-tex: From offline:

We want to make sure that callers cannot abuse this endpoint (on purpose or by accident), so we will limit this to a single execution at a time.

The best way to achieve this is by having this endpoint return immediately with a 202 Accepted and a response which points the caller to a `GET /sweep/status` endpoint which reports the status of the latest sweep - ongoing/finished.

F/U: Additionally, we want the status endpoint to return information about start and finish times for the sweep, number of unknown and missing skylinks handled, etc.

skyd/cache.go
ro-tex: Oh, you are right about the `struct{}` thing. Will change.

skyd/cache.go
ro-tex: Not currently but this will make the interface much more flexible, allowing us to add bulk operations more easily in the future. 👍

skyd/cache.go
ro-tex: This is just an implementation shortcut which I took while experimenting with the current implementation, similar to the complete lack of tests for this. I'll rework this to be part of the cache itself.

skyd/cache.go
ro-tex: Completely reworked.

api/handlers.go
```diff
@@ -89,12 +97,30 @@ func (api *API) unpinPOST(w http.ResponseWriter, req *http.Request, _ httprouter
...
+	// If there is no sweep in progress - kick one off.
+	if !api.latestSweepStatus.InProgress {
+		go api.threadedPerformSweep()
+	}
+	api.latestSweepStatusMu.Unlock()
+	// TODO If we want to be able to uniquely identify sweeps we can issue ids
+	//  for them and keep their statuses in a map. This would be the appropriate
+	//  RESTful approach. I am not sure we need that because all we care about
+	//  is to be able to kick off one and wait for it to end and this
+	//  implementations is sufficient for that.
```
ro-tex, line 115: I'm looking for opinions here, that's why the TODO is left in the code.

api/api.go
```diff
@@ -93,6 +106,24 @@ func (api *API) WriteJSON(w http.ResponseWriter, obj interface{}) {
...
+// header is set accordingly.
+func (api *API) WriteJSONCustomStatus(w http.ResponseWriter, obj interface{}, status int) {
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+	w.WriteHeader(status)
+	api.staticLogger.Traceln(status)
+	err := json.NewEncoder(w).Encode(obj)
+	if err != nil {
+		api.staticLogger.Debugln(err)
+	}
+	if _, isJSONErr := err.(*json.SyntaxError); isJSONErr {
```
MSevey, line 120: does this not panic if err is `nil`?

ro-tex: No, the type assertion only panics if you don't read its "ok" value, in this case `isJSONErr`. I believe using an assertion on a `nil` value works just like a method call on `nil` receiver - as long as the method doesn't access the receiver there is no issue. Playing around with this empirically proved those assumptions.

skyd/cache.go
```diff
@@ -0,0 +1,168 @@
...
+			Ch:        make(chan struct{}),
+			ExternErr: nil,
+		}
+		// Kick off the actual rebuild in a separate goroutine.
+		go psc.threadedRebuild(skydClient)
+	}
+	return *psc.result
+}
+
+// Remove registers the given skylinks in the cache.
```
MSevey, line 108:
```suggestion
// Remove removes the given skylinks in the cache.
```

skyd/cache.go
```diff
@@ -0,0 +1,168 @@
...
+func (psc *PinnedSkylinksCache) Remove(skylinks ...string) {
+	psc.mu.Lock()
+	defer psc.mu.Unlock()
+	for _, s := range skylinks {
+		delete(psc.skylinks, s)
+	}
+}
+
+// isRebuildInProgress returns true if a cache rebuild is in progress.
+// Calling this method assumes that caller is holding a lock on the cache.
```
MSevey, line 118: nit: this is implied by our converntions of this method being unexported and unmanaged.

skyd/cache.go
```diff
@@ -0,0 +1,168 @@
...
+		// Update the result.
+		psc.result.ExternErr = err
+		close(psc.result.Ch)
+		// Mark the rebuild as done.
+		psc.result = nil
+		psc.mu.Unlock()
+	}()
+
+	// Walk the entire Skynet folder and scan all files we find for skylinks.
+	dirsToWalk := []skymodules.SiaPath{skymodules.SkynetFolder}
```
MSevey, line 140: Follow Up: this is implemented in skyd now. You can call the client method `SkynetSkylinksGet()` to dedup this code.

skyd/cache.go
```diff
@@ -0,0 +1,168 @@
...
+// errors by setting the psc.err variable and it always closes the rebuildCh on
+// exit.
+func (psc *PinnedSkylinksCache) threadedRebuild(skydClient NodeSkydClient) {
+	var err error
+	// Ensure that we properly wrap up the rebuild process.
+	defer func() {
+		psc.mu.Lock()
+		// Update the result.
+		psc.result.ExternErr = err
+		close(psc.result.Ch)
```
MSevey, line 133: Follow Up: for these concurrent channels, we have implemented `Close()` functions on the struct in `skyd` so that we can check if the channel was already closed before closing it. This helps protect against panics for closing a closed channel.

skyd/cache.go
```diff
@@ -0,0 +1,168 @@
...
+		// Ch indicates the status of the cache rebuild progress - if it's not
+		// closed then the rebuild is still in progress.
+		Ch chan struct{}
+		// ExternErr holds the error state of the cache rebuild process. It must
+		// only be read after Ch is closed.
+		ExternErr error
+	}
+	// NodeSkydClient describes the part of skydclient.Client which we need.
+	// This type exists for testing purposes.
+	NodeSkydClient interface {
```
MSevey, line 32: I don't really like that we are making multiple skyd client interfaces throughout the code base.  If the reason for doing it here is to simply save a few lines of test code I'd prefer we didn't. 

Just my 2c, if I'm alone on this then that is ok.

api/handlers.go
```diff
@@ -117,3 +143,96 @@ func (api *API) parseAndResolve(skylink string) (skymodules.Skylink, error) {
...
+	unknown, missing := api.staticSkydClient.DiffPinnedSkylinks(dbSkylinks)
+
+	// Remove all unknown skylink from the database.
+	var s skymodules.Skylink
+	for _, sl := range unknown {
+		s, err = database.SkylinkFromString(sl)
+		if err != nil {
+			err = errors.AddContext(err, "invalid skylink found in DB")
+			build.Critical(err)
+			return
```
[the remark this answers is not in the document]
ro-tex: I agree. I was thinking "critical == stop what you are doing" but there isn't a need to do that.
I'll change those to `continue`s. I'll keep the other errors as `return`s because they will likely be caused by something more general like a DB failure.

api/handlers.go
ro-tex: Implementing this would require us to store those runs in the DB, so they would accessible from any server in the cluster.

Actually, we need to do that anyway for this monitoring to be able happen reliably because with the current approach the caller will get unreliable data if they get directed to another server by the LB. So, this definitely needs to live in the DB.

This should be a F/U, though. I would basically remove the endpoint until then or mark it as "only call this locally".

api/handlers.go
```diff
@@ -21,6 +25,10 @@ type (
 	SkylinkRequest struct {
 		Skylink string
 	}
+	// SweepPOSTResponse is the response to POST /sweep
+	SweepPOSTResponse struct {
+		Href string
```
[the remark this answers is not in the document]
ro-tex: I was following the example given [here](https://restfulapi.net/http-status-202-accepted/).

> although honestly I don't think this is necessary. developers have the API documentation to see how to get the status, they shouldn't have to rely on the response of one endpoint to know to call another.

That might be fine here where we don't refer to the specific request or the id of the current run. But the proper approach would be to provide a link which points to a specific sweep execution. I talk about that more in https://github.com/SkynetLabs/pinner/pull/22#discussion_r877152153 . 

Let's discuss this on Linear: https://linear.app/skynetlabs/issue/SKY-527/detailed-information-in-get-sweepstatus

skyd/cache.go
ro-tex: I know but I wanted to specifically spell it out. We are not very strict with this particular convention outside of skyd.

skyd/cache.go
ro-tex: Already on the list: https://github.com/SkynetLabs/pinner/issues/3

skyd/cache.go
ro-tex: We have this implemented only for files, it seems, and we typically just call `close`, e.g.
```go
close(uc.staticUploadCompletedChan)
```
But if you'd prefer that, I added a `close` method which checks whether the channel is closed before closing it, thus preventing double-closing. The method is unexported because I don't want it to acquire its own locks - I want the setting of the error and the closing of the channel to happen in that specific order and within the same lock scope.

skyd/cache.go
ro-tex: That's fair. To be honest, initially I thought those were two separate clients. 😅
Then I kept it because of some circular dependencies which forced me to put the mock next to the client and we typically don't like that at all. But in this case it makes some sense. I just went over it with Chris and we stared at it for 10 minutes trying to figure out a better approach and this seems to be the best compromise.

So, TL;DR:
 - I'll merge the two mocks
 - the common mock will be used by both the unit tests and the integration tests
 - the common mock will live in the main package next to the client implementation

MSevey (approved):

skyd/cache.go
```diff
@@ -0,0 +1,175 @@
...
+		result   *RebuildCacheResult
+		skylinks map[string]struct{}
+		mu       sync.Mutex
+	}
+	// RebuildCacheResult informs the caller on the status of a cache rebuild.
+	// The error should not be read before the channel is closed.
+	RebuildCacheResult struct {
+		// Ch indicates the status of the cache rebuild progress - if it's not
+		// closed then the rebuild is still in progress.
+		Ch chan struct{}
```
ChrisSchinnerl, line 26: (lord of the nits) f/u You could make this a `<-chan struct{}` to prevent the receiver of the result from closing the channel.
`ErrAvail` would be a nice name I think. I thought we usually call them `c` so I had to double check but apparently we commonly use nice names.

test/utils.go
```diff
@@ -90,3 +90,13 @@ func RandomSkylink() skymodules.Skylink {
 	sl, _ := skymodules.NewSkylinkV1(h, 0, 0)
 	return sl
 }
+
+// Contains checks whether the given slice contains the given element.
+func Contains[T comparable](haystack []T, needle T) bool {
```
ChrisSchinnerl, line 95: So it has finally come to this... C'est la vie simpler times.

test/tester.go
```diff
@@ -278,3 +283,20 @@ func (t *Tester) UnpinPOST(sl string) (int, error) {
 	r, err := t.Request(http.MethodPost, "/unpin", nil, body, nil, nil)
 	return r.StatusCode, err
 }
+
+// SweepPOST kicks off a background process which gets all files pinned by skyd
+// and marks them in the DB as pinned by the current server. It also goes over
+// all files in the DB that are marked as pinned by the local skyd and unmarks
+// those which are not in the list reported by skyd.
+func (t *Tester) SweepPOST() (api.SweepPOSTResponse, int, error) {
+	var resp api.SweepPOSTResponse
```
ChrisSchinnerl, line 292: For super short methods like these we sometimes use named return arguments. This is fine too though. Just a fun fact.

ChrisSchinnerl (approved):

PR #27, opened 2022-05-24
Post testing fixes

ro-tex:
# PULL REQUEST

## Overview

These are minor adjustments that I came up with while testing.

TL;DR:
 - when we get an non-breaking error (i.e. we continue processing other skylinks) while processing a skylink, we want to wait for a bit before moving on. this is both to avoid a tight loop around a faulty skylink as well giving more time for the cause of the error to go away, e.g. skyd to finish loading
 - slightly better logging

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/workers/scanner.go
+++ b/workers/scanner.go
@@ -314,7 +314,7 @@ func (s *Scanner) waitUntilHealthy(skylink skymodules.Skylink, sp skymodules.Sia
 		}
 		select {
 		case <-ticker.C:
-			s.staticLogger.Tracef("Waiting for '%s' to become fully healthy. Current health: %.2f", skylink, health)
+			s.staticLogger.Debugf("Waiting for '%s' to become fully healthy. Current health: %.2f", skylink, health)
 		case <-deadlineTimer.C:
 			s.staticLogger.Warnf("Skylink '%s' failed to reach full health within the time limit.", skylink)
 			break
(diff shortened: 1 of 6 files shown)


workers/scanner.go
```diff
@@ -171,12 +171,30 @@ func (s *Scanner) pinUnderpinnedSkylinks() {
...
-		// Block until the pinned skylink becomes healthy or until a timeout.
-		s.waitUntilHealthy(skylink, sp)
+		// We only check the error if we want to continue scanning. The error is
+		// already logged and the only indication it gives us is whether we
+		// should wait for the file we pinned to become healthy or not. If there
+		// is an error, then there is nothing to wait for.
+		if err == nil {
+			// Block until the pinned skylink becomes healthy or until a timeout.
+			s.waitUntilHealthy(skylink, sp)
+		} else {
```
MSevey, line 185: nit: I would put a `continue` after the `s.waitUntilHealthy` and then remove the else to have the select de-indented.

MSevey (dismissed):

workers/scanner.go
ro-tex: Might as well handle it right away. I used to opportunity to fix some typos and such as well.

MSevey (approved):

m-cat (approved):

PR #29, opened 2022-05-25
Make the sleepBetweenScans value configurable.

ro-tex:
# PULL REQUEST

## Checklist
 - [ ] Changelog file created


conf/configuration.go
```diff
@@ -131,6 +133,9 @@ func LoadConfig() (Config, error) {
 	if val, ok = os.LookupEnv("PINNER_LOG_LEVEL"); ok {
 		cfg.LogLevel = val
 	}
+	if val, ok = os.LookupEnv("PINNER_HOURS_BETWEEN_SCANS"); ok {
```
ChrisSchinnerl, line 136: I can't say why but I feel like hours is not something I'd expect when configuring a sleep.
I'd assume something like `PINNER_SLEEP_BETWEEN_SCANS` which by default accepts seconds and optionally allows for passing strings with suffixes like `24h`.

Makefile
```diff
@@ -130,14 +130,14 @@ bench: fmt
...
-test-long: lint lint-ci
+test-long: lint lint-ci start-mongo
 	@mkdir -p cover
 	GORACE='$(racevars)' go test -race --coverprofile='./cover/cover.out' -v -failfast -tags='testing debug netgo' -timeout=60s $(pkgs) -run=$(run) -count=$(count)
-
-test-int: test-long start-mongo
 	GORACE='$(racevars)' go test -race --coverprofile='./cover/cover.out' -v -tags='testing debug netgo' -timeout=600s $(integration-pkgs) -run=$(run) -count=$(count)
 	-make stop-mongo
 
+test-int: test-long
```
MSevey, line 139: same as other PR, we should just remove this.

conf/configuration.go
MSevey: agreed, I would prioritize consistency with other services here. Seconds is a much more common time interval.

workers/scanner.go
```diff
@@ -102,14 +98,25 @@ type (
 )
 
 // NewScanner creates a new Scanner instance.
-func NewScanner(db *database.DB, logger *logrus.Logger, minPinners int, serverName string, skydClient skyd.Client) *Scanner {
+func NewScanner(db *database.DB, logger *logrus.Logger, minPinners int, serverName string, customSleepBetweenScans string, skydClient skyd.Client) *Scanner {
+	sleep := sleepBetweenScans
+	if customSleepBetweenScans != "" {
+		hours, err := strconv.Atoi(customSleepBetweenScans)
+		if err != nil || hours < 1 {
+			logger.Warnf("Invalid value of PINNER_HOURS_BETWEEN_SCANS given, using the default of %d hours", sleepBetweenScans)
```
MSevey, line 106: I would return the error.  Setting invalid parameters should prevent a start up.

MSevey (changes requested):

conf/configuration.go
ro-tex: Added. It defaults to seconds when no suffix is given but also allows for time format such as `2h23m34s`.

workers/scanner.go
```diff
@@ -353,10 +363,12 @@ func (s *Scanner) waitUntilHealthy(skylink skymodules.Skylink, sp skymodules.Sia
...
 // SleepBetweenScans defines how often we'll scan the DB for underpinned
-// skylinks. The returned value varies by +/-sleepBetweenScansVariation and it's
+// skylinks. The returned value varies by +/-sleepVariationFactor and it's
 // centered on sleepBetweenScans.
-func SleepBetweenScans() time.Duration {
-	return sleepBetweenScans - sleepBetweenScansVariation + time.Duration(fastrand.Intn(2*int(sleepBetweenScansVariation)))
+func (s *Scanner) SleepBetweenScans() time.Duration {
+	sleep := float64(s.staticSleepBetweenScans)
+	sleepVar := sleep * sleepVariationFactor
+	return time.Duration(sleep-sleepVar) + time.Duration(fastrand.Intn(int(2*sleepVar)))
```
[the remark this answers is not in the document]
ro-tex: What would be a good name for the const? I am struggling to describe it succinctly.

Here's a different approach that removes that question entirely:
```go
	variation := int64(float64(s.staticSleepBetweenScans) * sleepVariationFactor)
	upper := int64(s.staticSleepBetweenScans) + variation
	lower := int64(s.staticSleepBetweenScans) - variation
	rng := upper - lower
	return time.Duration(rand.Int63n(rng) + lower)
```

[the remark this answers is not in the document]
ro-tex: I changed the code to that - it's more expressive.

MSevey (approved):

PR #30, opened 2022-05-25
Dry Run

ro-tex:
# PULL REQUEST

## Overview

Implement a dry_run cluster-level configuration option which omits all calls to skyd's pin/unpin.

The feature is a bit raw and underspecced, so feedback is welcome. We need to get it done ASAP, though, so we can kick this off in prod. We can also make this a temporary feature and remove it once the Pinner is stable and trusted.

Even if, for some reason, we decide to not go with this feature, there are changes in this PR that still need to be merged, e.g. `findAndPinOneUnderpinnedSkylink` being renamed to `managedFindAndPinOneUnderpinnedSkylink` and the lock-and-copy strat in there.

Internal discussion: https://discord.com/channels/542938080349519882/978547799799005184/979044946185183294

This depends on https://github.com/SkynetLabs/pinner/pull/27 because I needed some of the changes made there.
I'll select that branch as base for the moment, in order to make the reviews easier.

## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes SKY-676

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,10 +15,13 @@ all: release
 count = 1
 # pkgs changes which packages the makefile calls operate on. run changes which
 # tests are run during testing.
-pkgs = ./ ./api ./database ./skyd
+pkgs = ./ ./api ./database ./skyd ./test ./test/workers
 
 # integration-pkgs defines the packages which contain integration tests
 integration-pkgs = ./test ./test/api ./test/database
+
+# run determines which tests run when running any variation of 'make test'.
+run = .
 
 # fmt calls go fmt on all packages.
 fmt:
--- a/test/workers/scanner_test.go
+++ b/test/workers/scanner_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/skynetlabs/pinner/conf"
 	"github.com/skynetlabs/pinner/skyd"
 	"github.com/skynetlabs/pinner/test"
 	"github.com/skynetlabs/pinner/workers"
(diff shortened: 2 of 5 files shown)


MSevey (changes requested):
I think this looks good. We should be able to write a quick test that verifies in dry run a skylink doesn't get pinned. 

basically
```
trigger a pin event
wait the pin interval
verify skyd doesn't have the pin
turn off dry run
verify skyd gets the pin
```

m-cat (dismissed):

Makefile
```diff
@@ -15,11 +15,14 @@ all: release
 count = 1
 # pkgs changes which packages the makefile calls operate on. run changes which
 # tests are run during testing.
-pkgs = ./ ./api ./database ./skyd
+pkgs = ./ ./api ./database ./skyd ./test ./test/workers
```
m-cat, line 18: Not sure if I should have put the missing packages here  or in `integration-pkgs`.

test/workers/scanner_test.go
```diff
@@ -58,9 +59,100 @@ func TestScanner(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	// Wait for one cycle - the skylink should be picked up and pinned on the
-	// local skyd.
-	time.Sleep(workers.SleepBetweenScans())
+
+	// Wait - the skylink should be picked up and pinned on the local skyd.
+	time.Sleep(3 * workers.SleepBetweenScans())
```
m-cat, line 64: A single sleep-between-scans was not long enough for me...

m-cat (changes requested):
One possible issue I noticed with this yesterday (and somehow forgot about): if the scanner is running in dry mode, isn't it going to be picking up the same skylinks over and over, because they never get pinned? Should there be a new value like `dryPinned` or something to simulate marking them pinned?

MSevey:
> One possible issue I noticed with this yesterday (and somehow forgot about): if the scanner is running in dry mode, isn't it going to be picking up the same skylinks over and over, because they never get pinned? Should there be a new value like `dryPinned` or something to simulate marking them pinned?

yes they would get picked up over and over, that is fine, it is just going to be log spam, but ideally this would die down to only the skylinks that will need to be pinned.

one  of the reasons for the dry run mode was to reduce the risk of pinning a bunch of skylinks we don't need to as the cluster syncs up with the initial pinner deploy.

if we see a bunch of log spam, and then nothing, that would ultimate validate our concerns about deploying in dry run mode. 

So I don't think we should have any `dryPinned` value.

Makefile
MSevey: I would add them to `integration-pkgs` and then include `integration-pkgs` here. This is seemilar to how we handle the `lockcheck` packages in `skyd`.

fine as a follow up to clean up this file.

MSevey (approved):

m-cat (approved):
Looks good, waiting for @ro-tex to have a look and merge.

Makefile
ro-tex: https://github.com/SkynetLabs/pinner/issues/31

test/workers/scanner_test.go
```diff
@@ -58,9 +59,100 @@ func TestScanner(t *testing.T) {
...
+	err = db.SetConfigValue(ctx, conf.ConfDryRun, "false")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Verify skyd gets the pin.
+	//
+	// Wait enough time for the pinner to pick up the new value for dry_run and
+	// rescan.
+	time.Sleep(10 * workers.SleepBetweenScans())
```
ro-tex, line 154: This is way too long. We should replace it with `build.Retry`. We should actually replace all of these sleeps with that.

ro-tex: #31

PR #34, opened 2022-06-01
Fix test suite and prevent NDFs.

ro-tex:
# PULL REQUEST

## Overview

Some tests started randomly failing which is not great. This PR makes use of our tried and tested `build.Retry` mechanic and makes the tests much more reliable.

I also merged `test-long` and `test-int`. The difference between those used to be that `test-long` didn't need a test DB but the current ones do. So, I had a choice between adding them all in the `test-int` category and calling both groups. I went with the latter.

## Checklist

- [ ] Changelog file created.


test/workers/scanner_test.go
```diff
@@ -48,8 +56,9 @@ func TestScanner(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	// Sleep for two cycles.
-	time.Sleep(2 * workers.SleepBetweenScans())
+
+	// Sleep for a while, giving a chance to the scanner to pick the skylink up.
+	time.Sleep(cyclesToWait * workers.SleepBetweenScans())
```
ChrisSchinnerl, line 61: I would not use `workers.SleepBetweenScans` considering that it uses some randomness itself.
Instead I'd just define a `maxSleepBetweenScans` which is the largest possible outcome of `SleepBetweenScans` and sleep 2 times that. That should be plenty of time.

test/workers/scanner_test.go
```diff
@@ -119,8 +132,8 @@ func TestScannerDryRun(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	// Sleep for two cycles.
-	time.Sleep(2 * workers.SleepBetweenScans())
+	// Sleep for a while, giving a chance to the scanner to pick the skylink up.
+	time.Sleep(cyclesToWait * workers.SleepBetweenScans())
```
MSevey, line 136: same, we should be using build reties instead of bare sleeps if there are timing inconsistencies.

Makefile
```diff
@@ -130,14 +130,14 @@ bench: fmt
...
-test-long: lint lint-ci
+test-long: lint lint-ci start-mongo
 	@mkdir -p cover
 	GORACE='$(racevars)' go test -race --coverprofile='./cover/cover.out' -v -failfast -tags='testing debug netgo' -timeout=60s $(pkgs) -run=$(run) -count=$(count)
-
-test-int: test-long start-mongo
 	GORACE='$(racevars)' go test -race --coverprofile='./cover/cover.out' -v -tags='testing debug netgo' -timeout=600s $(integration-pkgs) -run=$(run) -count=$(count)
 	-make stop-mongo
 
+test-int: test-long
```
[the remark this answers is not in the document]
m-cat: If we do keep this, please leave a comment explaining the difference between `test-long` and `test-int`.

test/workers/scanner_test.go
ro-tex: Doing this in a retry would be misleading. When using retry we expect the state to change to the expected value at some point. In this case we want to ensure that no matter how long we wait the state will not change. That's why I'm waiting for the maximum allowed time for which the retry calls are running. This is similar to Chris's idea of `maxSleepBetweenScans` but it's also different because it's a lot longer. The shorter approach that he suggested just didn't cut it.

test/workers/scanner_test.go
ro-tex: Addressed in https://github.com/SkynetLabs/pinner/pull/34#discussion_r886854260

test/workers/scanner_test.go
```diff
@@ -132,7 +145,7 @@ func TestScannerDryRun(t *testing.T) {
 	}
 
 	// Wait - the skylink should not be picked up and pinned on the local skyd.
-	time.Sleep(3 * workers.SleepBetweenScans())
+	time.Sleep(cyclesToWait * workers.SleepBetweenScans())
```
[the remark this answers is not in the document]
ro-tex: Addressed in https://github.com/SkynetLabs/pinner/pull/34#discussion_r886854260

Makefile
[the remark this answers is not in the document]
m-cat: A short comment to that effect would satisfy me:

```
# Having this in place keeps the "testing interface" of the service compatible with accounts, blocker and so on.
```

m-cat (dismissed):
Looks good to me. 👍

Makefile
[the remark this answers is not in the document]
MSevey: Personally I would remove `test-int` in favor for more explicit options.

`test-long` is a standard, that is find. 

for tests that require mongo, we could either make a `test-long-mongo` or devs know which tests need to use the `start-mongo` command before running. 

or `test-long` allows runs mongo so it is never a concern. 

because if one of the goals with `test-int` is so you only need to run the integration services for the test that need the integration services, then that will be invalid if we add a second service, unless all those tests require both mongo and this new service. 

tl;dr, I think we are splitting hairs. 

From my perspective, I would rather the tests take a few seconds longer and have `test-long` always run mongo so that I never have to deal with determining which test command I should use for which tests.

Makefile
```diff
@@ -130,11 +130,13 @@ bench: fmt
 test:
 	go test -short -tags='debug testing netgo' -timeout=5s $(pkgs) -run=. -count=$(count)
 
+# Having this in place keeps the "testing interface" of the service compatible with accounts, blocker and so on.
 test-long: lint lint-ci
-	@mkdir -p cover
```
MSevey, line 134: Now it looks like `test-long` isn't running any tests which is very misleading.

ro-tex: Removed.

workers/scanner_test.go
```diff
@@ -1,14 +1,183 @@
...
+const (
+	// cyclesToWait establishes a common number of sleepBetweenScans cycles we
+	// should wait until we consider that a file has been or hasn't been picked
+	// by the scanner.
+	cyclesToWait = 5
+)
+
+var (
+	// maxSleepBetweenScans is the maximum time we might sleep between scans.
+	maxSleepBetweenScans = time.Duration(float64(sleepBetweenScans) * (1 + sleepVariationFactor))
```
ro-tex, line 25: The tests in this file were moved here because we no longer distinguish between integration and other long tests and it is better to have the tests closer and access unexported variables than exporting those variables for the sole reason that the tests need them.

MSevey (approved):

PR #35, opened 2022-06-20
Log to disk

ro-tex:
# PULL REQUEST

## Overview

This PR adds an optional env var which specifies a file on disk to which Pinner should write its logs. If the var is missing or empty, we don't write to disk.

## Checklist

- [ ] Changelog file created.

## Issues Closed

Closes SKY-756


MSevey (changes requested):

main.go
```diff
@@ -25,12 +27,11 @@ func main() {
...
 	ctx := context.Background()
-	logger := logrus.New()
-	logLevel, err := logrus.ParseLevel(cfg.LogLevel)
+	logger, loggerCloser, err := newLogger(cfg.LogLevel, cfg.LogFile)
 	if err != nil {
-		log.Fatal(errors.AddContext(err, "invalid log level: "+cfg.LogLevel))
+		log.Fatal(errors.AddContext(err, "failed to initialise logger"))
 	}
-	logger.SetLevel(logLevel)
+	defer loggerCloser()
```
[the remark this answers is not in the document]
ro-tex: It's returned by `newLogger()` on line 30.

I wouldn't push code that doesn't pass its tests and it can't pass its tests if it doesn't compile.

MSevey (approved):

conf/configuration.go
```diff
@@ -20,6 +20,7 @@ import (
 const (
 	defaultAccountsHost = "10.10.10.70"
 	defaultAccountsPort = "3000"
+	defaultLogFile      = "/logs/pinner.log"
```
ChrisSchinnerl, line 23: Don't you mean `./logs/pinner.log` to make it relative to where the pinner is running from?
If it's absolute, this won't run outside of docker without root privileges

ChrisSchinnerl: On the same note, can you add a test which makes sure that the log file is created at the default path and if overwritten by the environment var at the specified path?

ro-tex: The idea is for this to only work in docker, so yes, I mean the file at the root `/log` directory.
If somebody wants to run it outside of docker without root, they can pass the `PINNER_LOG_FILE` env var and set this value to whatever they want.

ro-tex: > On the same note, can you add a test which makes sure that the log file is created at the default path and if overwritten by the environment var at the specified path?

Sure. Do you mind if I do that in a F/U, so we don't need to wait for Matt to be back from vacation or involve another reviewer?

ro-tex: Here's the F/U: https://github.com/SkynetLabs/pinner/pull/38

ChrisSchinnerl: A f/u is fine if you have manually verified it's working. The goal is simply to verify this MR is doing what it claims to do. 1. log to disk 2. let you configure the log location.

Regarding the log location it's true that we ship using docker but we also want to run services locally for development and run tests locally without docker. Having the default experience be an error because the logfile can't be created is very annoying.

main.go
```diff
@@ -52,7 +53,45 @@ func main() {
...
+// stop using the logger, typically deferred in main.
+func newLogger(level, logfile string) (logger *logrus.Logger, closer func(), err error) {
+	logger = logrus.New()
+	// Parse and set log level.
+	logLevel, err := logrus.ParseLevel(level)
+	if err != nil {
+		return nil, nil, errors.AddContext(err, "invalid log level: "+level)
+	}
+	logger.SetLevel(logLevel)
+	// Open and start writing to the log file, unless we have an empty string,
```
ChrisSchinnerl, line 74: Empty strings are not great sentinel values. It's more common to pass an `io.Writer` to `newLogger` and then if you don't want to log you pass in `io.Discard`.

skyd/skyd.go
```diff
@@ -78,8 +78,8 @@ func (c *client) DiffPinnedSkylinks(skylinks []string) (unknown []string, missin
 // FileHealth returns the health of the given sia file.
 // Perfect health is 0.
 func (c *client) FileHealth(sp skymodules.SiaPath) (float64, error) {
-	c.staticLogger.Trace("Entering FileHealth.")
-	defer c.staticLogger.Trace("Exiting  FileHealth.")
+	c.staticLogger.Trace("Entering FileHealth")
```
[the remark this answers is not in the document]
ro-tex: > not sure how useful these trace logs are

They have proven to be very valuable in letting me verify that the service is running exactly how I planned it to. Observability. They are great.

> shouldn't we at least log the siapath?

Not really. I care about the execution path, not the specific paths which are being crawled. Logging the siapath will only add noise at this point.

main.go
peterjan: +1 was pondering commenting on this myself but didn't want to overly review it as I wasn't assigned

main.go
```diff
@@ -52,7 +53,45 @@ func main() {
...
+	logger.Printf("GitRevision: %v (built %v)", build.GitRevision, build.BuildTime)
 	err = server.ListenAndServe(4000)
 	log.Fatal(errors.Compose(err, scanner.Close()))
 }
+
+// newLogger creates a new logger that can write to disk.
+//
+// The function also returns a closer function that should be called when we
+// stop using the logger, typically deferred in main.
+func newLogger(level, logfile string) (logger *logrus.Logger, closer func(), err error) {
```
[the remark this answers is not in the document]
ro-tex: I thought about that. But having in mind that we'd never have anything else to do with it other than just log it, it felt like unnecessary complexity that I could neatly fold into a lower level.

main.go
```diff
@@ -25,12 +27,11 @@ func main() {
 	// the service. Once the context is closed, any background threads will
 	// wind themselves down.
 	ctx := context.Background()
-	logger := logrus.New()
-	logLevel, err := logrus.ParseLevel(cfg.LogLevel)
+	logger, loggerCloser, err := newLogger(cfg.LogLevel, cfg.LogFile)
```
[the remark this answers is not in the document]
ro-tex: Sure.

skyd/skyd.go
```diff
@@ -126,8 +126,8 @@ func (c *client) Pin(skylink string) (skymodules.SiaPath, error) {
 // on which the caller can either wait or select. The caller can check whether
 // the rebuild was successful by calling Error().
 func (c *client) RebuildCache() RebuildCacheResult {
-	c.staticLogger.Trace("Entering RebuildCache.")
-	defer c.staticLogger.Trace("Exiting  RebuildCache.")
+	c.staticLogger.Trace("Entering RebuildCache")
+	defer c.staticLogger.Trace("Exiting  RebuildCache")
```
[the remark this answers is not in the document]
ro-tex: It helps so much when reading the logs...

skyd/skyd.go
```diff
@@ -89,8 +89,8 @@ func (c *client) FileHealth(sp skymodules.SiaPath) (float64, error) {
 
 // Metadata returns the metadata of the skylink
 func (c *client) Metadata(skylink string) (skymodules.SkyfileMetadata, error) {
-	c.staticLogger.Trace("Entering Metadata.")
-	defer c.staticLogger.Trace("Exiting  Metadata.")
+	c.staticLogger.Trace("Entering Metadata")
+	defer c.staticLogger.Trace("Exiting  Metadata")
```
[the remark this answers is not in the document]
ro-tex: Same - it just makes my life better.

skyd/skyd.go
```diff
@@ -89,8 +89,8 @@ func (c *client) FileHealth(sp skymodules.SiaPath) (float64, error) {
 
 // Metadata returns the metadata of the skylink
 func (c *client) Metadata(skylink string) (skymodules.SkyfileMetadata, error) {
-	c.staticLogger.Trace("Entering Metadata.")
-	defer c.staticLogger.Trace("Exiting  Metadata.")
+	c.staticLogger.Trace("Entering Metadata")
```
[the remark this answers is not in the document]
ro-tex: Not really. It tells me the execution path. Same as the siapath in the other discussion, I don't particularly care which branch goes where - I won't debug what has happened with a particular skylink. I want to see that the service goes through the right paths.

Also, the code is not really concurrent - we have one thread that's doing the scanning. We don't want it concurrent because we want the other servers to also have a chance at pinning some of the underpinned skylinks.

workers/scanner.go
```diff
@@ -149,17 +149,17 @@ func (s *Scanner) threadedScanAndPin() {
 			}
 		}
 
-		s.staticLogger.Tracef("Start scanning.")
+		s.staticLogger.Tracef("Start scanning")
```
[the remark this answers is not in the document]
ro-tex: That's a great suggestion and I would have certainly gone for it if we had a way to tune the log level per some logical group. But since we don't, it will just result in more logspam. I've tried to hit a useful middle ground of logging just enough to know the execution path. This is basically the logging I used while developing and testing this (with some adjustments, obviously).

main.go
ro-tex: The entire reason for this function to exist is to be able to answer the question "was a log file defined?" and to be able to open a writer to it.

Would using a more explicit sentinel value (e.g. "DO_NOT_LOG_TO_FILE") make this better?

peterjan: I think the most common approach would be to add a `FILE_LOGGING_ENABLED` environment variable, perhaps in combination with a `FILE_LOGGING_PATH`. Or `LOG_ENABLED` and `LOG_PATH` or something along those lines that makes the most sense?

peterjan: Aha I see you already have `PINNER_LOG_FILE`, so maybe `PINNER_LOG_ENABLED`?

workers/scanner.go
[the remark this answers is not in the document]
peterjan: Resolving this because it doesn't really matter but I just meant adding 

```go
s.staticLogger.Tracef("Start scanning")
s.staticLogger.Tracef("1/3 managedRefreshDryRun...")
s.managedRefreshDryRun()
s.staticLogger.Tracef("2/3...")
...
```

main.go
```diff
@@ -52,7 +53,45 @@ func main() {
...
+	logger = logrus.New()
+	// Parse and set log level.
+	logLevel, err := logrus.ParseLevel(level)
+	if err != nil {
+		return nil, nil, errors.AddContext(err, "invalid log level: "+level)
+	}
+	logger.SetLevel(logLevel)
+	// Open and start writing to the log file, unless we have an empty string,
+	// which signifies "don't log to disk".
+	if logfile != "" {
```
peterjan, line 76: What happens when you specify a path to a directory that doesn't exist? It'll error out right?

peterjan (approved):

main.go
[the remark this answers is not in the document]
peterjan: Ok, even if when I don't really need it I usually just write `closeFn func() error` because you expect an error from a close function 🤷‍♂️ . But it doesn't matter much I guess it's just PP

ChrisSchinnerl (approved):

conf/configuration.go
ro-tex: Our current situation is "never log to disk". This PR changes that to "allow logging to disk on demand". I don't see a reason to jump all the way to "always log to disk".

main.go
ro-tex: F/U: https://github.com/SkynetLabs/pinner/issues/39

PR #38, opened 2022-07-01
Test LoadConfig and Logger

ro-tex:
# PULL REQUEST

## Overview

Adds some tests as F/U for https://github.com/SkynetLabs/pinner/pull/35

## Checklist

- [ ] Changelog file created.

## Issues Closed

Closes SKY-756

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -28,10 +28,6 @@ func NewLogger(level logrus.Level, logfile string) (logger *logrus.Logger, close
 		// Create a closer function which flushes the content to disk and closes
 		// the log file gracefully.
 		closer = func() {
-			if e := fh.Sync(); e != nil {
-				log.Println(errors.AddContext(err, "failed to sync log file to disk"))
-				return
-			}
 			if e := fh.Close(); e != nil {
 				log.Println(errors.AddContext(err, "failed to close log file"))
 				return
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -13,14 +13,9 @@ import (
 //
 // The function also returns a closer function that should be called when we
 // stop using the logger, typically deferred in main.
-func NewLogger(level, logfile string) (logger *logrus.Logger, closer func(), err error) {
+func NewLogger(level logrus.Level, logfile string) (logger *logrus.Logger, closer func(), err error) {
 	logger = logrus.New()
-	// Parse and set log level.
-	logLevel, err := logrus.ParseLevel(level)
-	if err != nil {
-		return nil, nil, errors.AddContext(err, "invalid log level: "+level)
-	}
-	logger.SetLevel(logLevel)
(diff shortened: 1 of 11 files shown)


logger/logger.go
```diff
@@ -0,0 +1,47 @@
...
+		var fh *os.File
+		fh, err = os.OpenFile(logfile, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0644)
+		if err != nil {
+			return nil, nil, errors.AddContext(err, "failed to open log file")
+		}
+		logger.SetOutput(io.MultiWriter(os.Stdout, fh))
+		// Create a closer function which flushes the content to disk and closes
+		// the log file gracefully.
+		closer = func() {
+			if e := fh.Sync(); e != nil {
```
ChrisSchinnerl, line 36: You can keep this but just FYI it's not necessary.

logger/logger.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// NewLogger creates a new logger that can write to disk.
+//
+// The function also returns a closer function that should be called when we
+// stop using the logger, typically deferred in main.
+func NewLogger(level, logfile string) (logger *logrus.Logger, closer func(), err error) {
+	logger = logrus.New()
+	// Parse and set log level.
+	logLevel, err := logrus.ParseLevel(level)
```
ChrisSchinnerl, line 19: This still feels like it would belong into `LoadConfig` tbh.  There is not reason to not make the LogLevel field of the Config a `logrus.LogLevel` and immediately parse the environment var.
The sooner we turn vague user input into strong types the better.

logger/logger.go
```diff
@@ -0,0 +1,47 @@
...
+	if logfile != "" {
+		var fh *os.File
+		fh, err = os.OpenFile(logfile, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0644)
+		if err != nil {
+			return nil, nil, errors.AddContext(err, "failed to open log file")
+		}
+		logger.SetOutput(io.MultiWriter(os.Stdout, fh))
+		// Create a closer function which flushes the content to disk and closes
+		// the log file gracefully.
+		closer = func() {
```
ChrisSchinnerl, line 35: The more common way to do this would be passing the file in and handling the defer outside of this function and passing it in as an `io.Writer` as I mentioned in the other PR already (to also get rid of the sentinel value) or to wrap the logger in your own type and implement `Close` on it which you then defer in the outer scope.
That avoids allocating the function on the heap.

ChrisSchinnerl (approved):

logger/logger.go
ro-tex: Implemented a wrapper type.

peterjan (approved):

ChrisSchinnerl (approved):

PR #40, opened 2022-07-07
Implement a sweeping schedule.

ro-tex:
# PULL REQUEST

## Overview

Implement a sweeping schedule. This will help Pinner stay up to date with all files pinned by the servers, even if nginx fails to notify it. This is only a backup measure but it's rather useful.

The sweeps are very slow if there is a lot to sweep but they are very quick if there isn't. I don't expect this to put any measurable additional pressure on the servers.

## Checklist

- [ ] Changelog file created.

## Issues Closed
Closes SKY-1101

### DIFF
--- a/database/skylink.go
+++ b/database/skylink.go
@@ -113,20 +113,20 @@ func (db *DB) MarkUnpinned(ctx context.Context, skylink skymodules.Skylink) erro
 	return err
 }
 
-// AddServerForSkylink adds a new server to the list of servers known to be pinning
-// this skylink. If the skylink does not already exist in the database it will
-// be inserted. This operation is idempotent.
+// AddServerForSkylinks adds a new server to the list of servers known to be
+// pinning these skylinks. If a skylink does not already exist in the database
+// it will be inserted. This operation is idempotent.
 //
-// The `markPinned` flag sets the `unpin` field of the skylink to false when
+// The `markPinned` flag sets the `unpin` field of a skylink to false when
 // raised but it doesn't set it to false when not raised. The reason for that is
 // that it accommodates a specific use case - adding a server to the list of
 // pinners of a given skylink will set the unpin field to false is we are doing
 // that because we know that a user is pinning it but not so if we are running
 // a server sweep and documenting which skylinks are pinned by this server.
-func (db *DB) AddServerForSkylink(ctx context.Context, skylink skymodules.Skylink, server string, markPinned bool) error {
-	db.staticLogger.Tracef("Entering AddServerForSkylink. Skylink: '%s', server: '%s'", skylink, server)
-	defer db.staticLogger.Tracef("Exiting  AddServerForSkylink. Skylink: '%s', server: '%s'", skylink, server)
-	filter := bson.M{"skylink": skylink.String()}
+func (db *DB) AddServerForSkylinks(ctx context.Context, skylinks []string, server string, markPinned bool) error {
+	db.staticLogger.Tracef("Entering AddServerForSkylinks. Skylink: '%v', server: '%s'", skylinks, server)
+	defer db.staticLogger.Tracef("Exiting  AddServerForSkylinks. Skylink: '%v', server: '%s'", skylinks, server)
+	filter := bson.M{"skylink": bson.M{"$in": skylinks}}
 	var update bson.M
 	if markPinned {
 		update = bson.M{
@@ -137,22 +137,22 @@ func (db *DB) AddServerForSkylink(ctx context.Context, skylink skymodules.Skylin
 		update = bson.M{"$addToSet": bson.M{"servers": server}}
 	}
 	opts := options.Update().SetUpsert(true)
-	_, err := db.staticDB.Collection(collSkylinks).UpdateOne(ctx, filter, update, opts)
+	_, err := db.staticDB.Collection(collSkylinks).UpdateMany(ctx, filter, update, opts)
 	return err
 }
 
-// RemoveServerFromSkylink removes a server to the list of servers known to be
-// pinning this skylink. If the skylink does not exist in the database it will
-// not be inserted.
-func (db *DB) RemoveServerFromSkylink(ctx context.Context, skylink skymodules.Skylink, server string) error {
-	db.staticLogger.Tracef("Entering RemoveServerFromSkylink. Skylink: '%s', server: '%s'", skylink, server)
-	defer db.staticLogger.Tracef("Exiting  RemoveServerFromSkylink. Skylink: '%s', server: '%s'", skylink, server)
+// RemoveServerFromSkylinks removes a server from the list of servers known to
+// be pinning these skylinks. If a skylink does not exist in the database it
+// will not be inserted.
+func (db *DB) RemoveServerFromSkylinks(ctx context.Context, skylinks []string, server string) error {
+	db.staticLogger.Tracef("Entering RemoveServerFromSkylinks. Skylink: '%v', server: '%s'", skylinks, server)
+	defer db.staticLogger.Tracef("Exiting  RemoveServerFromSkylinks. Skylink: '%v', server: '%s'", skylinks, server)
 	filter := bson.M{
-		"skylink": skylink.String(),
+		"skylink": bson.M{"$in": skylinks},
 		"servers": server,
 	}
 	update := bson.M{"$pull": bson.M{"servers": server}}
-	_, err := db.staticDB.Collection(collSkylinks).UpdateOne(ctx, filter, update)
+	_, err := db.staticDB.Collection(collSkylinks).UpdateMany(ctx, filter, update)
 	return err
 }
 
--- a/main.go
+++ b/main.go
@@ -50,7 +50,7 @@ func main() {
 	}
 	swpr := sweeper.New(db, skydClient, cfg.ServerName)
 	// Schedule a sweep every 24 hours.
-	swpr.Schedule(24 * time.Hour)
+	swpr.UpdateSchedule(24 * time.Hour)
 
 	// Initialise the server.
 	server, err := api.New(cfg.ServerName, db, logger, skydClient, swpr)
--- a/main.go
+++ b/main.go
@@ -52,7 +52,7 @@ func main() {
 	}
 	swpr := sweeper.New(db, skydClient, cfg.ServerName)
 	// Schedule a sweep every 24 hours.
-	swpr.Schedule().Update(24*time.Hour, swpr)
+	swpr.UpdateSchedule(24 * time.Hour)
 
 	// Initialise the server.
 	server, err := api.New(cfg.ServerName, db, logger, skydClient, swpr)
--- a/main.go
+++ b/main.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"log"
 	"os"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	"github.com/skynetlabs/pinner/api"
@@ -12,6 +13,7 @@ import (
 	"github.com/skynetlabs/pinner/conf"
 	"github.com/skynetlabs/pinner/database"
 	"github.com/skynetlabs/pinner/skyd"
+	"github.com/skynetlabs/pinner/sweeper"
 	"github.com/skynetlabs/pinner/workers"
 	"gitlab.com/NebulousLabs/errors"
 )
@@ -46,9 +48,12 @@ func main() {
 	if err != nil {
 		log.Fatal(errors.AddContext(err, "failed to start Scanner"))
 	}
+	swpr := sweeper.New(db, skydClient, cfg.ServerName)
+	// Schedule a sweep every 24 hours.
+	swpr.Schedule(24 * time.Hour)
 
 	// Initialise the server.
-	server, err := api.New(cfg.ServerName, db, logger, skydClient)
+	server, err := api.New(cfg.ServerName, db, logger, skydClient, swpr)
 	if err != nil {
 		log.Fatal(errors.AddContext(err, "failed to build the api"))
 	}
--- a/main.go
+++ b/main.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"log"
-	"time"
 
 	"github.com/skynetlabs/pinner/api"
 	"github.com/skynetlabs/pinner/build"
@@ -50,9 +49,9 @@ func main() {
 	if err != nil {
 		log.Fatal(errors.AddContext(err, "failed to start Scanner"))
 	}
-	swpr := sweeper.New(db, skydClient, cfg.ServerName)
-	// Schedule a sweep every 24 hours.
-	swpr.UpdateSchedule(24 * time.Hour)
+	swpr := sweeper.New(db, skydClient, cfg.ServerName, logger)
+	// Schedule a regular sweep..
+	swpr.UpdateSchedule(sweeper.SweepInterval)
 
 	// Initialise the server.
 	server, err := api.New(cfg.ServerName, db, logger, skydClient, swpr)
--- a/main.go
+++ b/main.go
@@ -52,7 +52,7 @@ func main() {
 	}
 	swpr := sweeper.New(db, skydClient, cfg.ServerName)
 	// Schedule a sweep every 24 hours.
-	swpr.UpdateSchedule(24 * time.Hour)
+	swpr.Schedule().Update(24*time.Hour, swpr)
 
 	// Initialise the server.
 	server, err := api.New(cfg.ServerName, db, logger, skydClient, swpr)
--- a/sweeper/sweeper.go
+++ b/sweeper/sweeper.go
@@ -2,7 +2,6 @@ package sweeper
 
 import (
 	"context"
-	"sync"
 	"time"
 
 	"github.com/skynetlabs/pinner/database"
@@ -17,19 +16,6 @@ const (
 )
 
 type (
-	// Status represents the status of a sweep.
-	Status struct {
-		InProgress bool
-		Error      error
-		StartTime  time.Time
-		EndTime    time.Time
-	}
-	// status is the internal type we use when we want to be able to modify it.
-	status struct {
-		Status
-		mu           sync.Mutex
-		staticLogger logger.ExtFieldLogger
-	}
 	// Sweeper takes care of sweeping the files pinned by the local skyd server
 	// and marks them as pinned by the local server.
 	Sweeper struct {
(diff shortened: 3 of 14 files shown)


sweeper/sweeper.go
```diff
@@ -0,0 +1,188 @@
...
+//  for them and keep their statuses in a map. This would be the appropriate
+//  RESTful approach. I am not sure we need that because all we care about
+//  is to be able to kick off one and wait for it to end and this
+//  implementations is sufficient for that.
+func (s *Sweeper) Sweep() {
+	go s.threadedPerformSweep()
+}
+
+// threadedPerformSweep performs the actual sweep operation.
+func (s *Sweeper) threadedPerformSweep() {
```
ro-tex, line 101: This method was directly moved from the `api` package without any changes in logic. Feel free to not review.

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+	defer s.scheduleMu.Unlock()
+	if s.schedule != nil {
+		close(s.schedule.cancelCh)
+	}
+	s.schedule = &Schedule{
+		Period:   period,
+		cancelCh: make(chan struct{}),
+	}
+	go func() {
+		t := time.NewTicker(s.schedule.Period)
```
ChrisSchinnerl, line 64: You need to stop this ticker to avoid leaking resources

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+func New(db *database.DB, skydc skyd.Client, serverName string) *Sweeper {
+	return &Sweeper{
+		staticDB:         db,
+		staticSkydClient: skydc,
+		staticServerName: serverName,
+	}
+}
+
+// Schedule a scan to run on each period.
+func (s *Sweeper) Schedule(period time.Duration) {
```
ChrisSchinnerl, line 53: My preference would be to call this `UpdateSchedule` since simply `Update` makes it sound like you are scheduling just 1 sweep.

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+	s.scheduleMu.Lock()
+	defer s.scheduleMu.Unlock()
+	if s.schedule != nil {
+		close(s.schedule.cancelCh)
+	}
+	s.schedule = &Schedule{
+		Period:   period,
+		cancelCh: make(chan struct{}),
+	}
+	go func() {
```
ChrisSchinnerl, line 63: This could be a separate `threadedSweepLoop` function.

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+		staticServerName: serverName,
+	}
+}
+
+// Schedule a scan to run on each period.
+func (s *Sweeper) Schedule(period time.Duration) {
+	s.scheduleMu.Lock()
+	defer s.scheduleMu.Unlock()
+	if s.schedule != nil {
+		close(s.schedule.cancelCh)
```
ChrisSchinnerl, line 57: If you want this to be very accurate you can add a `sync.WaitGroup` to the schedule.
Then, after closing this channel, you wait on the waitgroup for the existing goroutine to be done.

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+				s.Sweep()
+			case <-s.schedule.cancelCh:
+				return
+			}
+		}
+	}()
+}
+
+// Scheduled returns a copy of the current Sweeper schedule.
+func (s *Sweeper) Scheduled() Schedule {
```
ChrisSchinnerl, line 77: `Schedule` or `ActiveSchedule`. `Scheduled` sounds like the result would be a `bool`.

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+	if s.schedule == nil {
+		return Schedule{}
```
ChrisSchinnerl, line 79: Why not return a pointer? Checking for `nil` is nicer and faster than a deep comparison against `Schedule{}`.

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	// Sweeper takes care of sweeping the files pinned by the local skyd server
+	// and marks them as pinned by the local server.
+	Sweeper struct {
+		staticDB         *database.DB
+		staticSkydClient skyd.Client
+		staticServerName string
+
+		schedule   *Schedule
+		scheduleMu sync.Mutex
```
peterjan, line 36: If you have a mutex guarding a nested struct, you can make it `static` and push the mutex into the struct, which makes more sense because it's only guarding those fields anyway. In `skyd` we have the `call` and `managed` conventions to go alongside it so you and up with 

```go
sweeper.staticSchedule.callUpdatePeriod(newPeriod)
```
instead of 
```go
sweeper.scheduleMu.Lock()
sweeper.schedule.period = newPeriod
sweeper.scheduleMu.Unlock()
```

which is nice because you push the locking behaviour and the management into the sub struct so you don't have to care about it in your parent struct, you only have to pay attention that you're not holding two locks at the same time.

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+	"gitlab.com/NebulousLabs/errors"
+	"gitlab.com/SkynetLabs/skyd/build"
+	"gitlab.com/SkynetLabs/skyd/skymodules"
+)
+
+type (
+	// Schedule defines how often, if at all, we sweep this server automatically.
+	Schedule struct {
+		Period   time.Duration
+		cancelCh chan struct{}
```
peterjan, line 19: `nit` `cancelChan`?

sweeper/sweeper.go
ChrisSchinnerl: Also is this unused?

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+		Error:      nil,
+		StartTime:  time.Now().UTC(),
+		EndTime:    time.Time{},
+	}
+	s.statusMu.Unlock()
+	// Define an error variable which will represent the success of the scan.
+	var err error
+	// Ensure that we'll finalize the sweep on returning from this method.
+	defer func() {
+		s.statusMu.Lock()
```
peterjan, line 123: would be better if all of this stuff is defined as methods on the `status` struct, that way you can really easily write unit tests for it as well

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+
+// Sweep starts a new skyd sweep, unless one is already underway.
+//
+// TODO If we want to be able to uniquely identify sweeps we can issue ids
+//  for them and keep their statuses in a map. This would be the appropriate
+//  RESTful approach. I am not sure we need that because all we care about
+//  is to be able to kick off one and wait for it to end and this
+//  implementations is sufficient for that.
+func (s *Sweeper) Sweep() {
+	go s.threadedPerformSweep()
```
peterjan, line 100: should we consider checking the status here to avoid launching the goroutine?

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+// threadedPerformSweep performs the actual sweep operation.
+func (s *Sweeper) threadedPerformSweep() {
+	s.statusMu.Lock()
+	// Double-check for parallel sweeps.
+	if s.status.InProgress {
+		s.statusMu.Unlock()
+		return
+	}
+	// Initialise the status to "a sweep is running".
+	s.status = Status{
```
peterjan, line 112: this could be extracted into a constructor `newStatus` and the whole thing could be `callInitSweep` on the status that returns a `bool` indicating if one started, in case it returns `false` you could return

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	wg.Wait()
+	if cacheErr != nil {
+		err = errors.AddContext(cacheErr, "failed to rebuild skyd cache")
+		return
+	}
+
+	unknown, missing := s.staticSkydClient.DiffPinnedSkylinks(dbSkylinks)
+
+	// Remove all unknown skylink from the database.
```
peterjan, line 162:
```suggestion
	// Remove all unknown skylinks from the database.
```

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+		skylink, err = database.SkylinkFromString(sl)
+		if err != nil {
+			err = errors.AddContext(err, "invalid skylink found in DB")
+			build.Critical(err)
+			continue
+		}
+		err = s.staticDB.RemoveServerFromSkylink(ctx, skylink, s.staticServerName)
+		if err != nil {
+			err = errors.AddContext(err, "failed to unpin skylink")
+			return
```
peterjan, line 174: wouldn't we want to try all of them instead of returning on failure?

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+	// Remove all unknown skylink from the database.
+	var skylink skymodules.Skylink
+	for _, sl := range unknown {
+		skylink, err = database.SkylinkFromString(sl)
+		if err != nil {
+			err = errors.AddContext(err, "invalid skylink found in DB")
+			build.Critical(err)
+			continue
+		}
+		err = s.staticDB.RemoveServerFromSkylink(ctx, skylink, s.staticServerName)
```
peterjan, line 171: can we perform these operations in bulk? usually doing mongo writes/updates in a for loop is a flag for me, if we can I think we should optimise this because it's so much slower to perform every update individually

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	// Add all missing skylinks to the database.
+	for _, sl := range missing {
+		skylink, err = database.SkylinkFromString(sl)
+		if err != nil {
+			err = errors.AddContext(err, "invalid skylink reported by skyd")
+			build.Critical(err)
+			continue
+		}
+		err = s.staticDB.AddServerForSkylink(ctx, skylink, s.staticServerName, false)
```
peterjan, line 185: can we perform these operations in bulk? usually doing mongo writes/updates in a for loop is a flag for me, if we can I think we should optimise this because it's so much slower to perform every update individually

sweeper/sweeper.go
```diff
@@ -0,0 +1,191 @@
...
+		if err != nil {
+			err = errors.AddContext(err, "failed to unpin skylink")
+			return
```
peterjan, line 188: wouldn't we want to try all of them instead of returning on failure?  
also, update the error message here I think it's incorrect

peterjan (changes requested):
there's no unit tests? this definitely warrants unit tests and ideally also an integration test to verify the handlers

sweeper/sweeper.go
ro-tex: That sounds odd to me but if it sounds better to you then I'm fine with it.

sweeper/sweeper.go
ro-tex: I'm not sure that's a good idea. On one hand, we're not reusing this anywhere and it's a very short and straightforward function. On the other hand, it interacts with the `s.schedule` struct which we set up here. I'm not sure I'd like to even make it possible for this logic to be reused.

What is the benefit which you have in mind for moving this to a named function?

sweeper/sweeper.go
ro-tex: > `Scheduled` sounds like the result would be a `bool`.

This is so counter-intuitive to me. :( In my mind boolean functions are `IsScheduled` and such. And "schedule" is both a noun and a verb, that's why `Schedule(period)` sounded as a good constructor name.

> Also is this unused?

Right, maybe we don't need it, yet. I'll remove it for the moment.

sweeper/sweeper.go
ro-tex: The logic here was that if we fail to add a server that means we have a Mongo issue and it wouldn't make sense to continue. If we continue with the other skylinks, we'll just spam the logs with potentially dozens of thousands of lines.

The message is wrong, thanks. Fixed.

sweeper/sweeper.go
ro-tex: Good point! Yes, we absolutely should turn this into a bulk update.

sweeper/sweeper.go
```diff
@@ -0,0 +1,140 @@
...
+//  RESTful approach. I am not sure we need that because all we care about
+//  is to be able to kick off one and wait for it to end and this
+//  implementations is sufficient for that.
+func (s *Sweeper) Sweep() {
+	go s.threadedPerformSweep()
+}
+
+// UpdateSchedule schedules a new series of sweeps to be run.
+// If there are already scheduled sweeps, that schedule is cancelled (running
+// // sweeps are not interrupted) and a new schedule is established.
```
MSevey, line 65:
```suggestion
// sweeps are not interrupted) and a new schedule is established.
```

sweeper/sweeper.go
```diff
@@ -0,0 +1,140 @@
...
+	// Double-check for parallel sweeps.
+	if s.status.InProgress {
+		s.statusMu.Unlock()
+		return
+	}
+	// Initialise the status to "a sweep is running".
+	s.status = Status{
+		InProgress: true,
+		Error:      nil,
+		StartTime:  time.Now().UTC(),
```
MSevey, line 82: nit: Why do we need the `UTC()`? We just need the server to be consistent right? If a server is not in `UTC` timezone this is going to be inaccurate.

MSevey: If we don't need it, it would make is safer since we wouldn't be reliant on a developer remembering to add the `UTC()` when getting the current time.

MSevey (approved):

sweeper/sweeper.go
ro-tex: I took a second look here and we don't need a goroutine at all. The cache rebuild happens in its own goroutine anyway.

sweeper/sweeper.go
ro-tex: Done.

sweeper/sweeper.go
ro-tex: I went with the Start/Finalize approach and a static status. Otherwise I would have needed two mutexes to protect a single field and that would have been confusing. Now Start initialises a sweep (unless one is already in progress) and Finalize finishes it.

database/skylink.go
```diff
@@ -113,20 +113,20 @@ func (db *DB) MarkUnpinned(ctx context.Context, skylink skymodules.Skylink) erro
...
 // raised but it doesn't set it to false when not raised. The reason for that is
 // that it accommodates a specific use case - adding a server to the list of
 // pinners of a given skylink will set the unpin field to false is we are doing
 // that because we know that a user is pinning it but not so if we are running
 // a server sweep and documenting which skylinks are pinned by this server.
-func (db *DB) AddServerForSkylink(ctx context.Context, skylink skymodules.Skylink, server string, markPinned bool) error {
-	db.staticLogger.Tracef("Entering AddServerForSkylink. Skylink: '%s', server: '%s'", skylink, server)
-	defer db.staticLogger.Tracef("Exiting  AddServerForSkylink. Skylink: '%s', server: '%s'", skylink, server)
-	filter := bson.M{"skylink": skylink.String()}
+func (db *DB) AddServerForSkylinks(ctx context.Context, skylinks []string, server string, markPinned bool) error {
```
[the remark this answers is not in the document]
ro-tex: I generally agree with the idea and I would have kept these as `skymodules.Skylink` if I could. But the way we call this method is like this:
```go
	unknown, missing := s.staticSkydClient.DiffPinnedSkylinks(dbSkylinks)
	// Add all missing skylinks to the database.
	err = s.staticDB.AddServerForSkylinks(ctx, missing, s.staticServerName, false)
```
and `DiffPinnedSkylinks` operates with strings because it works with data coming from the database. So, if we decide that this is important we can convert the strings to skylinks somewhere for validation purposes but I'd rather do it at the points where those enter the system, i.e. in the sweeper and in the handlers. We are already doing it in the handlers and we're not doing it while sweeping because we trust the contents of the files in the renter's dir. I can definitely add a check there, if that would be helpful.

main.go
```diff
@@ -48,9 +50,12 @@ func main() {
 	if err != nil {
 		log.Fatal(errors.AddContext(err, "failed to start Scanner"))
 	}
+	swpr := sweeper.New(db, skydClient, cfg.ServerName)
+	// Schedule a sweep every 24 hours.
+	swpr.UpdateSchedule(24 * time.Hour)
```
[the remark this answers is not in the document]
ro-tex: Done.

sweeper/sweeper.go
```diff
@@ -0,0 +1,140 @@
...
+func (s *Sweeper) Status() Status {
+	s.statusMu.Lock()
+	st := s.status
+	s.statusMu.Unlock()
+	return st
+}
+
+// Sweep starts a new skyd sweep, unless one is already underway.
+//
+// TODO If we want to be able to uniquely identify sweeps we can issue ids
```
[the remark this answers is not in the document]
ro-tex: Added some logging.

sweeper/sweeper.go
ro-tex: It's really helpful to have all of our servers report the same time zone. This is one less thing to worry about, especially when we write things to the DB.

If we discover that this is a problem and people are forgetting about this we should add our own Time type which does it for us.

database/skylink.go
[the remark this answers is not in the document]
MSevey: a simple check that all the strings are valid siapaths could be helpful. That would catch data corruption on disk or some other misconfiguration, like somehow looking at the wrong directory. But like you said, maybe in the handler and not here. 

either way, fine for a follow up, or if it is more effort then help we can just dismiss this.

sweeper/sweeper.go
MSevey: ok, I'd add a quick comment then so that we remember the reasoning in the future. I'd put the comment on the `StartTime` in the `Status` struct. 

fine as a follow up.

MSevey (approved):

database/skylink.go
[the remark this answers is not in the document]
ro-tex: I can add it i #41.

sweeper/sweeper.go
ro-tex: Done in #41.

sweeper/sweeper.go
ChrisSchinnerl: The benefit is mostly readability and documentability (not sure that is a word lol) but it's not a dealbreaker for this PR.

ChrisSchinnerl (approved):

peterjan (approved):

sweeper/sweeper.go
peterjan: ok

PR #44, opened 2022-08-03
Mark server as dead/removed

ro-tex:
# PULL REQUEST

## Overview

This PR adds a new endpoint which removes a server from the cluster, a.k.a. marks it as dead. This causes the server to be removed as pinner from all skylink it was previously pinning. Also, a new scan is scheduled for an hour later, so we can start repinning all those freshly underpinned skylinks on other servers.

## Checklist

- [ ] Changelog file created.

## Issues Closed

Closes SKY-297

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -3,6 +3,7 @@ package api
 import (
 	"encoding/json"
 	"net/http"
+	"time"
 
 	"github.com/julienschmidt/httprouter"
 	"github.com/skynetlabs/pinner/conf"
@@ -12,6 +13,16 @@ import (
 )
 
 type (
+	// ServerRemoveRequest describes a payload that marks a server as dead.
+	ServerRemoveRequest struct {
+		Server string `json:"server"`
+	}
+	// ServerRemoveResponse returns the removed server and the number of
+	// skylinks it was pinning.
+	ServerRemoveResponse struct {
+		Server      string `json:"server"`
+		NumSkylinks int64  `json:"numSkylinks"`
+	}
 	// HealthGET is the response type of GET /health
 	HealthGET struct {
 		DBAlive    bool `json:"dbAlive"`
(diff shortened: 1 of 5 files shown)


api/handlers.go
```diff
@@ -95,6 +106,41 @@ func (api *API) unpinPOST(w http.ResponseWriter, req *http.Request, _ httprouter
...
+		api.WriteError(w, errors.New("no server found in request body"), http.StatusBadRequest)
+		return
+	}
+	// Remove the server as pinner.
+	n, err := api.staticDB.RemoveServer(req.Context(), body.Server)
+	if err != nil {
+		api.WriteError(w, errors.AddContext(err, "failed to remove server"), http.StatusInternalServerError)
+		return
+	}
+	// Schedule a scan for underpinned skylinks in an hour, so all of them can
```
peterjan, line 128: `nit` maybe move this up before the server removal? I think that's safer because if scheduling fails the server won't be removed, and if the server's removed but scheduling fails, that's annoying.

peterjan (approved):

api/handlers.go
ro-tex: Sounds good. I'll do it in a F/U, so we don't need to re-approve. I'll keep this open until then, so I don't forget.

ChrisSchinnerl: Quickly hijacking this to ask a question. If a server is removed and we schedule a scan for 1 hour in the future. Then we remove another one 30 minutes later. Will the scan happen at the original time or 1 hour from the second removal?

ChrisSchinnerl (approved):

api/handlers.go
ro-tex: This is a good point. Servers check for the next scan each 30 minutes until the next scan is scheduled for less than 30 minutes in the future. When a server sees that a scan is scheduled in less than 30 minutes it sleeps its scanner until the exact moment of the scan and performs it without checking again. I did that in order to avoid endless checking and some complications caused by clock desynchronisation between servers.

So, if two servers are checking for next scan and server1 checks 35 minutes before the scan it will have to check again 5 minutes before the scan again. if server2 checks 25 minutes before the scan it will not check again. Then, if we delete a server 20 minutes before the scan then a new scan will be scheduled for one hour later. server1 will check and will wait for that hour before scanning. Server2 will not - it will sleep for its 25 minutes and then do a scan.

I believe the best approach here is to check for next scan before scheduling one and if it's in less than an hour just not schedule one because we've got it coming anyway.

Good catch! I'll implement it in #45 as F/U.

PR #45, opened 2022-08-04
Limit repinning underpinned skylinks to underutilized servers

ro-tex:
# PULL REQUEST

## Overview

The only servers which would automatically repin underpinned skylinks:
 - have less than 50TiB in contract data
 - are the server in the cluster with least data
 - are in the bottom 30% of all servers in the cluster by pinned data

## Checklist

- [ ] Changelog file created.

## Issues Closed
Closes SKY-1078

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"encoding/json"
+	"github.com/skynetlabs/pinner/lib"
 	"net/http"
 	"time"
 
@@ -123,7 +124,7 @@ func (api *API) serverRemovePOST(w http.ResponseWriter, req *http.Request, _ htt
 	// Schedule a scan for underpinned skylinks in an hour (unless one is
 	// already pending), so all of them can be repinned ASAP but also all
 	// servers in the cluster will have enough time to get the memo for the scan.
-	t := time.Now().UTC().Add(time.Hour)
+	t := lib.Now().Add(time.Hour)
 	t0, err := conf.NextScan(ctx, api.staticDB, api.staticLogger)
 	// We just set it when we encounter an error because we can get such an
 	// error in two cases - there is no next scan scheduled or there is a
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -2,15 +2,14 @@ package api
 
 import (
 	"encoding/json"
-	"github.com/skynetlabs/pinner/lib"
-	"net/http"
-	"time"
-
 	"github.com/julienschmidt/httprouter"
 	"github.com/skynetlabs/pinner/conf"
 	"github.com/skynetlabs/pinner/database"
+	"github.com/skynetlabs/pinner/lib"
+	"github.com/skynetlabs/pinner/workers"
 	"gitlab.com/NebulousLabs/errors"
 	"gitlab.com/SkynetLabs/skyd/skymodules"
+	"net/http"
 )
 
 type (
@@ -124,15 +123,16 @@ func (api *API) serverRemovePOST(w http.ResponseWriter, req *http.Request, _ htt
 	// Schedule a scan for underpinned skylinks in an hour (unless one is
 	// already pending), so all of them can be repinned ASAP but also all
 	// servers in the cluster will have enough time to get the memo for the scan.
-	t := lib.Now().Add(time.Hour)
+	t := lib.Now().Add(workers.SleepBeforeForcedScan)
 	t0, err := conf.NextScan(ctx, api.staticDB, api.staticLogger)
 	// We just set it when we encounter an error because we can get such an
 	// error in two cases - there is no next scan scheduled or there is a
 	// problem with the DB. In the first case we want to schedule one and in the
 	// second we'll get the error again with the next operation.
 	if err != nil || t0.After(t) {
-		err = conf.SetNextScan(ctx, api.staticDB, t)
+		err1 := conf.SetNextScan(ctx, api.staticDB, t)
 		if err != nil {
+			err = errors.Compose(err1, errors.AddContext(err, "failed to fetch next scan"))
 			api.WriteError(w, errors.AddContext(err, "failed to schedule a scan"), http.StatusInternalServerError)
 			return
 		}
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -2,14 +2,27 @@ package api
 
 import (
 	"encoding/json"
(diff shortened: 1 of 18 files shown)


ro-tex:
I'm still chasing some deadlock in the tests but the PR can be reviewed.

api/handlers.go
```diff
@@ -119,19 +119,33 @@ func (api *API) serverRemovePOST(w http.ResponseWriter, req *http.Request, _ htt
...
+	// already pending), so all of them can be repinned ASAP but also all
+	// servers in the cluster will have enough time to get the memo for the scan.
+	t := time.Now().UTC().Add(time.Hour)
+	t0, err := conf.NextScan(ctx, api.staticDB, api.staticLogger)
+	// We just set it when we encounter an error because we can get such an
+	// error in two cases - there is no next scan scheduled or there is a
+	// problem with the DB. In the first case we want to schedule one and in the
+	// second we'll get the error again with the next operation.
+	if err != nil || t0.After(t) {
+		err = conf.SetNextScan(ctx, api.staticDB, t)
```
ChrisSchinnerl, line 133: If you use `err = errors.Compose(err, conf.SetNextScan(...)` here you won't hide the previous error in case it's not nil.

api/handlers.go
```diff
@@ -119,19 +119,33 @@ func (api *API) serverRemovePOST(w http.ResponseWriter, req *http.Request, _ htt
 		api.WriteError(w, errors.New("no server found in request body"), http.StatusBadRequest)
 		return
 	}
+	ctx := req.Context()
+	// Schedule a scan for underpinned skylinks in an hour (unless one is
+	// already pending), so all of them can be repinned ASAP but also all
+	// servers in the cluster will have enough time to get the memo for the scan.
+	t := time.Now().UTC().Add(time.Hour)
```
ChrisSchinnerl, line 126: (nit) could be a constant

workers/scanner.go
```diff
@@ -43,6 +44,19 @@ const (
 )
 
 var (
+	// AlwaysPinThreshold sets a limit on the contract data of the server. If
+	// the server is below that limit, it will repin underpinned files eve if it
```
ChrisSchinnerl, line 48:
```suggestion
	// the server is below that limit, it will repin underpinned files even if it
```

workers/scanner.go
```diff
@@ -399,6 +425,36 @@ func (s *Scanner) managedRefreshMinPinners() {
 	s.mu.Unlock()
 }
 
+// staticEligibleToPin returns true when this server is either pinning less than
+// AlwaysPinThreshold data or it's in the lower PinningRangeThresholdPercent
+// section of servers when ordered by contract data in descending order.
+// A server is always eligible if it's last in the list.
+func (s *Scanner) staticEligibleToPin(ctx context.Context) (bool, error) {
+	pinnedData, err := s.staticDB.ServerLoad(ctx, s.staticServerName)
```
ChrisSchinnerl, line 433: tip: You already wrote a good test so there is no action item here but I would probably have written a `isEligibleToPin(pinnecData, pos, total)` function which is called by `scanner.staticEligibleToPin`.
Unit testing a function like that is super easy to do with a table test and then you only need to write a quick test for the scanner method which verifies that it calls the function.

workers/scanner.go
```diff
@@ -399,6 +425,36 @@ func (s *Scanner) managedRefreshMinPinners() {
...
+	if err != nil {
+		return false, err
+	}
+	// Below the hard limit.
+	if pinnedData < int64(AlwaysPinThreshold) {
+		return true, nil
+	}
+	pos, total, err := s.staticDB.ServerLoadPosition(ctx, s.staticServerName)
```
ChrisSchinnerl, line 441: (nit) if this also returned the pinned data, then you can fetch all the information you need from the database with a single read vs 2 interactions.

workers/scanner.go
```diff
@@ -399,6 +425,36 @@ func (s *Scanner) managedRefreshMinPinners() {
...
+	if err != nil {
+		return false, err
+	}
+	// Last in the list.
+	if pos == total {
+		return true, nil
+	}
+	// In the bottom PinningRangeThresholdPercent.
+	posPercent := 1.0 - float64(pos)/float64(total)
+	thresholdPercent := float64(PinningRangeThresholdPercent) / 100.0
```
ChrisSchinnerl, line 451: nit/tip: you could make the constant `0.3` instead of 30 and dividing by 100.0 to avoid the casting and the floating point division. Doesn't matter too much for this but constants in Go are weird cause they have infinite precision and depending on the value can be assigned to multiple different types without casting. Also more compile time checks. TLDR; constants are nice.

workers/scanner.go
```diff
@@ -399,6 +425,36 @@ func (s *Scanner) managedRefreshMinPinners() {
...
+	if posPercent < thresholdPercent {
+		return true, nil
+	}
+	return false, nil
```
ChrisSchinnerl, line 455: In case you feel fancy `return posPercent < thresholdPercent, nil`

ChrisSchinnerl (commented):
Lgtm. I left a few nits you could sneak in witht your deadlock fix if you want.

api/handlers.go
ro-tex: Done.

workers/scanner.go
ro-tex: Even the spell check can't save me from those ones. :D

workers/scanner.go
ro-tex: Good point.

workers/scanner.go
ro-tex: *facepalm*
Yes... Yes, that makes a lot of sense. :)
I won't change this one due to time pressure but that's a great tip. Thanks!

workers/scanner.go
ro-tex: You mean that I wouldn't need `ServerLoad`? Hm... On one hand one RT to the DB is better than two. On the other hand `ServerLoadPositions` returns all records in the table and also orders them by a field, which is a lot of work.

With the size of the table in question it's a toss-up between the two. I'll keep the status quo for the moment.

workers/scanner.go
ro-tex: > constants are nice.

Yes, very true.

Added. :)

ChrisSchinnerl (approved):

PR #52, opened 2022-08-11
Better health endpoint

ro-tex:
# PULL REQUEST

## Checklist

- [ ] Changelog file created.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -51,7 +51,9 @@ type (
 
 // healthGET returns the status of the service
 func (api *API) healthGET(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-	var status HealthGET
+	status := HealthGET{
+		DBAlive: true,
+	}
 	err := api.staticDB.Ping(req.Context())
 	if err != nil {
 		status.DBAlive = false
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -24,7 +24,8 @@ type (
 		Server      string `json:"server"`
 		NumSkylinks int64  `json:"numSkylinks"`
 	}
-	// HealthGET is the response type of GET /health
+	// HealthGET is the response type of GET /health.
+	// Primary field is only populated on error.
 	HealthGET struct {
 		DBAlive    bool   `json:"dbAlive"`
 		Error      string `json:"error"`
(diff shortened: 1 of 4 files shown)


api/handlers.go
```diff
@@ -41,9 +52,47 @@ type (
...
+		return
+	}
+	extHealth.Hello = hello
+
 	mp, err := conf.MinPinners(req.Context(), api.staticDB)
-	var status HealthGET
-	status.DBAlive = err == nil
+	if err != nil {
+		status.Error = err.Error()
+		status.Primary = hello.Primary
```
kwypchlo, line 92: this looks like primary will only be available if ping and hello succeeded - let's add it in case we get any error

api/handlers.go
```diff
@@ -41,9 +52,47 @@ type (
...
+			api.staticLogger.Warnf("Failed to serialize extended health information. Error: %v", err)
+		}
+		api.staticLogger.Info(string(b))
+	}()
+	err := api.staticDB.Ping(req.Context())
+	if err != nil {
+		status.DBAlive = false
+		status.Error = err.Error()
+		api.WriteJSON(w, status)
+		return
```
kwypchlo, line 79: we should still run other checks and return their status even if ping fails

kwypchlo (changes requested):

api/handlers.go
ro-tex: If the `ping` doesn't succeed then we don't have a live connection to the DB and we can't run the `hello`. If we can't run the `hello`, we can't find the primary.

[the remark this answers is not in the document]
ro-tex: Knowing the primary is definitely useful. It's also very useful to attackers because it tells them where to focus their efforts. That's why we decided to limit the reporting of the primary to error states only. The primary is always logged, so an operator will be able to get that info from the logs.

api/handlers.go
ro-tex: See my other response (https://github.com/SkynetLabs/pinner/pull/52/files#r943655859).

api/handlers.go
kwypchlo: that's not entirely true since we've seen a case where ping would fail on one service but reading health check succeeded on the other - those do not seem to be mutually exclusive, probably because ping has 5s timeout and read doesn't so read can succeed even if ping failed right?

ro-tex: right, ping does have a short timeout. maybe this is something we should change as well?

ro-tex: Anyway, I changed the logic to always try to get all the information it possibly can. This might often result in multiple repeated errors in the error string but it will be more informative than what we had previously.

MSevey (approved):

kwypchlo (approved):

ChrisSchinnerl (approved):

api/handlers.go
```diff
@@ -41,10 +52,48 @@ type (
...
+	}
+
 	mp, err := conf.MinPinners(req.Context(), api.staticDB)
-	var status HealthGET
-	status.DBAlive = err == nil
-	status.MinPinners = mp
+	if err != nil {
+		status.Error = errors.Compose(status.Error, err)
+		if hello != nil {
+			status.Primary = hello.Primary
```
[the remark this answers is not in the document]
ro-tex: Done in a F/U.

PR #53, opened 2022-08-12
Skip blocked skylinks

ro-tex:
# PULL REQUEST

## Overview

This PR adds two things:
1. Do not process skylinks which are blocked by `skyd`.
2. Do not process skylinks which have already caused an error during the current scan.

## Checklist

- [ ] Changelog file created.

### DIFF
--- a/skyd/cache.go
+++ b/skyd/cache.go
@@ -1,8 +1,8 @@
 package skyd
 
 import (
-	"bytes"
 	"fmt"
+	"go.sia.tech/siad/crypto"
 	"sync"
 
 	"gitlab.com/NebulousLabs/errors"
@@ -93,13 +93,13 @@ func (psc *PinnedSkylinksCache) Diff(sls []string) (unknown []string, missing []
 // rebuilding happens in a goroutine, allowing the method to return a channel
 // on which the caller can either wait or select. The caller can check whether
 // the rebuild was successful by calling Error().
-func (psc *PinnedSkylinksCache) Rebuild(skydClient Client, omitBlockedSkylinks bool) RebuildCacheResult {
+func (psc *PinnedSkylinksCache) Rebuild(skydClient Client) RebuildCacheResult {
 	psc.mu.Lock()
 	defer psc.mu.Unlock()
 	if !psc.isRebuildInProgress() {
 		psc.result = NewRebuildCacheResult()
 		// Kick off the actual rebuild in a separate goroutine.
-		go psc.threadedRebuild(skydClient, omitBlockedSkylinks)
+		go psc.threadedRebuild(skydClient)
 	}
 	return *psc.result
 }
@@ -122,7 +122,7 @@ func (psc *PinnedSkylinksCache) isRebuildInProgress() bool {
 // threadedRebuild performs the actual cache rebuild process. It reports any
 // errors by setting the psc.err variable and it always closes the rebuildCh on
 // exit.
-func (psc *PinnedSkylinksCache) threadedRebuild(skydClient Client, omitBlockedSkylinks bool) {
+func (psc *PinnedSkylinksCache) threadedRebuild(skydClient Client) {
 	var err error
 	// Ensure that we properly wrap up the rebuild process.
 	defer func() {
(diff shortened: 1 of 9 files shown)


skyd/cache.go
```diff
@@ -163,6 +164,33 @@ func (psc *PinnedSkylinksCache) threadedRebuild(skydClient Client) {
...
+			return
+		}
+		var sl skymodules.Skylink
+		for s := range sls {
+			err = sl.LoadString(s)
+			if err != nil {
+				// This is an invalid skylink, remove it.
+				delete(sls, s)
+			}
+			for _, bl := range blocklist.Blocklist {
```
ChrisSchinnerl, line 183: Not sure how big the blocklist is at this point but it might be worthwhile to turn this into a map first.

skyd/cache.go
```diff
@@ -163,6 +164,33 @@ func (psc *PinnedSkylinksCache) threadedRebuild(skydClient Client) {
...
+		if err != nil {
+			err = errors.AddContext(err, "failed to fetch blocklist")
+			return
+		}
+		var sl skymodules.Skylink
+		for s := range sls {
+			err = sl.LoadString(s)
+			if err != nil {
+				// This is an invalid skylink, remove it.
+				delete(sls, s)
```
ChrisSchinnerl, line 181: Shouldn't you `continue` here?

sweeper/sweeper.go
```diff
@@ -83,7 +83,7 @@ func (s *Sweeper) threadedPerformSweep() {
 	// Kick off a skyd client cache rebuild. That happens in a separate
 	// goroutine. We'll block on the result channel only after we're done with
 	// the other tasks we can do while waiting.
-	res := s.staticSkydClient.RebuildCache()
+	res := s.staticSkydClient.RebuildCache(true)
```
ChrisSchinnerl, line 86: If we always call it with `true` we might not need the argument?

workers/scanner.go
```diff
@@ -327,7 +328,12 @@ func (s *Scanner) managedFindAndPinOneUnderpinnedSkylink() (skylink skymodules.S
 	minPinners := s.minPinners
 	s.mu.Unlock()
 
-	sl, err := s.staticDB.FindAndLockUnderpinned(context.TODO(), s.staticServerName, minPinners)
+	ctx := context.TODO()
+	// skipSkylinks is a list of skylinks which we want to skip during this
+	// scan. These might be skylinks which errored out or blocked skylinks.
+	skipSkylinks := make([]string, 0)
```
ChrisSchinnerl, line 334: Could be `var skipSkylinks []string` I think. For loops and `len` should still handle it correctly.

skyd/cache.go
ro-tex: Oh, that's true. This might take quite a while otherwise.

skyd/cache.go
ro-tex: Absolutely. Thanks!

sweeper/sweeper.go
ro-tex: I initially started without the argument. But then I thought that it's quite limiting to always exclude those blocked skylinks. Maybe we'll need the cache for something else as well like listing all the files pinned by `skyd`? Maybe it's a case of "you ain't gonna need it". I'll remove the argument - it's easy to add later if we need it.

workers/scanner.go
ro-tex: Sure. I wanted to be sure that `FindAndLockUnderpinned` will be fine using it because I don't really trust the mongo driver to handle nil slices correctly. But looking at this again, I should make handle that in `FindAndLockUnderpinned` instead.

Changed.

workers/scanner.go
```diff
@@ -327,7 +328,12 @@ func (s *Scanner) managedFindAndPinOneUnderpinnedSkylink() (skylink skymodules.S
 	minPinners := s.minPinners
 	s.mu.Unlock()
 
-	sl, err := s.staticDB.FindAndLockUnderpinned(context.TODO(), s.staticServerName, minPinners)
+	ctx := context.TODO()
+
+	// skipSkylinks is a list of skylinks which we want to skip during this
+	// scan. These might be skylinks which errored out or blocked skylinks.
+	var skipSkylinks []string
```
MSevey, line 335: I might just be missing it but I don't see how this is being used. 

You initialize it here and immediately call `FindAndLockUnderpinned` with it, but it is empty at that point. 

You then add skylinks to it, but you never call out to another function with it or make a DB call to update skipped links. So how does the server know about these skipped links the next time around?

ro-tex: Right, good catch. Distraction strikes again -I thought i was putting it in the longer running upper method.

I'll move it to the scanner object itself and I'll wipe it on each scan start.

workers/scanner.go
```diff
@@ -279,6 +290,27 @@ func (s *Scanner) staticScheduleNextScan() bool {
...
 	defer s.staticLogger.Trace("Exiting  managedPinUnderpinnedSkylinks")
+
+	// Clear out the skipped skylinks from the previous run.
+	s.mu.Lock()
+	s.skipSkylinks = []string{}
+	s.mu.Unlock()
+
+	intermediateStatsTicker := time.NewTicker(printPinningStatisticsPeriod)
+	defer intermediateStatsTicker.Stop()
+	countPinned := 0
```
MSevey, line 301: nit: might be helpful to also count total or failed so that we know a little more about the performance of pinner and each statement is more informative in isolation.

MSevey (dismissed):

skyd/cache.go
```diff
@@ -163,6 +164,32 @@ func (psc *PinnedSkylinksCache) threadedRebuild(skydClient Client) {
...
+	blockMap := make(map[crypto.Hash]struct{}, len(blocklist.Blocklist))
+	for _, bl := range blocklist.Blocklist {
+		blockMap[bl] = struct{}{}
+	}
+	var sl skymodules.Skylink
+	for s := range sls {
+		err = sl.LoadString(s)
+		if err != nil {
+			// This is an invalid skylink, remove it.
+			delete(sls, s)
```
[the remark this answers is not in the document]
ro-tex: Reading and validating all skylinks twice didn't make much sense, so I reorganised the code a bit so can filter the blocked ones while we are walking the filesystem.

workers/scanner.go
ro-tex: That's hard to pull off right now with the current setup. I can add more fields directly on the scanner struct but that doesn't seem like a good approach. The good news is that all skylinks which fail are added to the skipped array. I'll add the number of skipped skylinks to the output and that will give us more info.

ChrisSchinnerl (approved):

ChrisSchinnerl (approved):

MSevey (approved):

MSevey (approved):

PR #58, opened 2022-08-30
Make the stats loop actually loop.

ro-tex:
# PULL REQUEST

## Overview

Make the stats loop actually loop.
Reset failed attempts on a successful server pin.

## Checklist

- [ ] Changelog file created.

### DIFF
--- a/database/skylink.go
+++ b/database/skylink.go
@@ -178,7 +178,10 @@ func (db *DB) AddServerForSkylinks(ctx context.Context, skylinks []string, serve
 	if markPinned {
 		update = bson.M{
 			"$addToSet": bson.M{"servers": server},
-			"$set":      bson.M{"pinned": true},
+			"$set": bson.M{
+				"pinned":          true,
+				"failed_attempts": 0,
+			},
 		}
 	} else {
 		update = bson.M{"$addToSet": bson.M{"servers": server}}
--- a/database/skylink.go
+++ b/database/skylink.go
@@ -184,7 +184,12 @@ func (db *DB) AddServerForSkylinks(ctx context.Context, skylinks []string, serve
 			},
 		}
 	} else {
-		update = bson.M{"$addToSet": bson.M{"servers": server}}
+		update = bson.M{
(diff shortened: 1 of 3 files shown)


database/skylink.go
```diff
@@ -178,7 +178,10 @@ func (db *DB) AddServerForSkylinks(ctx context.Context, skylinks []string, serve
 	if markPinned {
 		update = bson.M{
 			"$addToSet": bson.M{"servers": server},
-			"$set":      bson.M{"pinned": true},
+			"$set": bson.M{
+				"pinned":          true,
+				"failed_attempts": 0,
```
ChrisSchinnerl, line 183: This seems to scream for a regression test.

workers/scanner.go
```diff
@@ -282,29 +282,31 @@ func (s *Scanner) threadedPrintStats(stopCh chan struct{}) {
...
-		skipped := s.skipSkylinks
-		startTime := s.scanStart
-		s.mu.Unlock()
-		s.staticLogger.Infof("Finished at %s, runtime %s, pinned skylinks %d, skipped skylinks %d",
-			t1.Format(conf.TimeFormat), t1.Sub(startTime).String(), atomic.LoadUint32(&s.atomicCountPinned), len(skipped))
-		s.staticLogger.Tracef("Skipped %d skylinks: %v", len(skipped), skipped)
-	case <-s.staticTG.StopChan():
-		s.staticLogger.Trace("Stop channel closed")
-		return
+	for {
```
ChrisSchinnerl, line 285: (nit) FYI you can loop over channels.

e.g.

```go
for range intermediateStatsTicker.C {
  select {
    // check for interruption
  }
  // do stuff
}
```

ro-tex: On one hand, that's super neat I had forgotten about it, so thanks for reminding me.

On the other hand, that would block on the ticker and will only check for an interruption once the ticker ticks, which is once per 10 minutes. That's not what we want here.

ChrisSchinnerl (approved):

PR #62, opened 2022-09-09
Add a scan status endpoint

ro-tex:
# PULL REQUEST

## Checklist

- [ ] Changelog file created.


scanner/scanner_test.go
```diff
@@ -1,26 +1,27 @@
-package workers
+package scanner_test
```
ro-tex, line 1: No functional changes.

The only thing changed in this file is the new test package name which is used in order to work around a cyclical dependency with the test package.

scanner/scanner.go
```diff
@@ -1,20 +1,19 @@
-package workers
+package scanner
```
ro-tex, line 1: The only changed lines in this file are the ones that contain "status". Everything else is the same, just moved around. I'm sorry for the large and complex diff but it really needed some house keeping before we wrap the service up.

api/handlers.go
```diff
@@ -263,6 +264,17 @@ func (api *API) sweepStatusGET(w http.ResponseWriter, _ *http.Request, _ httprou
 	api.WriteJSON(w, api.staticSweeper.Status())
 }
 
+// scannerStatusGET responds with the status of the latest scan.
+func (api *API) scannerStatusGET(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	verbose, _ := strconv.ParseBool(r.FormValue("verbose"))
```
ChrisSchinnerl, line 269: I'm usually in favour of forcing the parsing to succeed if an optional flag is specified to help users immediately see potential mistakes they made on their end. e.g. a typo

MSevey (approved):

ChrisSchinnerl (approved):

api/handlers.go
ro-tex: Yes, I generally agree with that as well. In this case it felt unnecessary but maybe I'm just being lazy. I'll add it with the next PR.

